PR #34, opened 2024-12-09
feat: Implement Gift Exchange Members

### LINKED ISSUE #33: [Group] GroupExchange Members Endpoints
Build API endpoints for gift exchange members management

Endpoints:

```
POST /api/gift-exchanges/:exchangeId/members     // Add member to exchange
GET /api/gift-exchanges/:exchangeId/members      // Get all members of an exchange
PATCH /api/gift-exchanges/:exchangeId/members/:memberId   // Update member status/recipient
DELETE /api/gift-exchanges/:exchangeId/members/:memberId  // Remove member from exchange
GET /api/users/:userId/gift-exchanges            // Get all exchanges a user is part of
```

thomasnguyen:
closes #33

This PR takes care of the members <--> gift exchange functionality. This includes a member joining a giftexchange.

All of the supabase models have been created

### DIFF
--- a/app/api/gift-exchanges/[id]/members/[memberId]/route.ts
+++ b/app/api/gift-exchanges/[id]/members/[memberId]/route.ts
@@ -1,0 +1,84 @@
+import { createClient } from "@/lib/supabase/server";
+import { NextResponse } from "next/server";
+import type { UpdateGiftExchangeMemberRequest } from "@/app/types/giftExchangeMember";
+
+// update a gift exchange member
+export async function PATCH(req: Request) {
+  const { searchParams } = new URL(req.url);
+  const exchangeId = searchParams.get("exchangeId");
+  const memberId = searchParams.get("memberId");
+
+  try {
+    const supabase = await createClient();
+    const {
+      data: { user },
+    } = await supabase.auth.getUser();
+
+    if (!user) {
+      return NextResponse.json({ error: "Unauthorized" }, { status: 401 });
+    }
+
+    const body: UpdateGiftExchangeMemberRequest = await req.json();
+
+    const { data, error } = await supabase
+      .from("gift_exchange_members")
+      .update({
+        ...body,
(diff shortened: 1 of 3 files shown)


app/api/gift-exchanges/[id]/members/[memberId]/route.ts
```diff
@@ -0,0 +1,84 @@
...
+      return NextResponse.json({ error: error.message }, { status: 500 });
+    }
+
+    return NextResponse.json({ success: true });
+  } catch (error) {
+    console.log(error);
+
+    return NextResponse.json(
+      { error: "Internal server error" },
+      { status: 500 }
```
Comment (jennifertieu) [inline app/api/gift-exchanges/[id]/members/[memberId]/route.ts:81]: Question: Whats the reason why we're explicitly setting the error response here?

Comment (thomasnguyen) [reply app/api/gift-exchanges/[id]/members/[memberId]/route.ts:81]: Good question! So this error handling is in the catch block. Its just a fallback if an unexpected error occurred.

If the deletion is succesful, we would return true
`return NextResponse.json({ success: true });`

Comment (jennifertieu) [reply app/api/gift-exchanges/[id]/members/[memberId]/route.ts:81]: I see, thank you Thomas. So does Supabase error object not return a status code (`error.status`) or its just preferable to explicitly set it to `500`? I just want to understand what your thought process was here for error handling.

Comment (thomasnguyen) [reply app/api/gift-exchanges/[id]/members/[memberId]/route.ts:81]: Ah I see what you're saying. So we generally dont want to expose the raw Supabase error details. Generally its not good practice as it could potentially reveal sensitive information and is a security concern.

5xx error code signifies that its a backend error code vs a frontend

Comment (jennifertieu) [reply app/api/gift-exchanges/[id]/members/[memberId]/route.ts:81]: I got it. That makes sense and is good to know!! Thank you for explaining, Thomas! I'm going to keep that in mind for the future.

jennifertieu (approved):
LGTM

PR #55, opened 2024-12-18
Jennifer/populate profile page

### LINKED ISSUE #40: [Profile] Fetch and populate profile into page
Fetch the user's profile and populate page

Image

If the profile page is the users' show edit button

jennifertieu:
Closes #40 

* Fetches current user profile information and populates profile
* Fetches other user profile information and populates their profile

### DIFF
--- a/app/profile/LoadingCard.tsx
+++ b/app/profile/LoadingCard.tsx
@@ -1,0 +1,25 @@
+import { Skeleton } from "@/components/Skeleton/Skeleton";
+
+const LoadingCard = () => {
+  return (
+    <article className="text-white px-4">
+      <div className="flex items-center rounded-t-2xl py-5 px-9 bg-groupCardGreen">
+        <Skeleton className="w-full h-10 rounded-full" />
+      </div>
+      <div className="bg-[#12433A] rounded-b-2xl grid grid-cols-1 lg:grid-cols-[minmax(550px,2fr)_minmax(300px,1fr)] gap-12 pl-9 pr-28 py-7">
+        <div className="flex flex-col gap-5">
+          <Skeleton className="w-full h-10 rounded-md" />
+          <Skeleton className="w-full h-10 rounded-md" />
+          <Skeleton className="w-full h-10 rounded-md" />
+        </div>
+        <div className="flex flex-col gap-5">
+          <Skeleton className="w-full h-10 rounded-md" />
+          <Skeleton className="w-full h-10 rounded-md" />
+          <Skeleton className="w-full h-10 rounded-md" />
+        </div>
+      </div>
+    </article>
+  );
+};
+
+export default LoadingCard;
--- a/app/types/profile.ts
+++ b/app/types/profile.ts
@@ -1,19 +1,19 @@
 export interface Profile {
-	id: string;
-	name: string;
-	email: string;
-	age_group?: string;
-	categories: string[];
-	hobbies?: string;
-	avoid?: string;
-	practical_whimsical?: number;
-	cozy_adventurous?: number;
-	minimal_luxurious?: number;
-	created_at: string;
-	updated_at: string;
-	onboarding_complete: boolean;
+  id: string;
+  display_name: string;
+  email: string;
+  age_group?: string;
+  categories: string[];
+  hobbies?: string;
+  avoid?: string;
+  practical_whimsical?: number;
+  cozy_adventurous?: number;
+  minimal_luxurious?: number;
+  created_at: string;
+  updated_at: string;
+  onboarding_complete: boolean;
 }
 
 export type ProfileUpdate = Partial<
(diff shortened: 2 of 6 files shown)


thomasnguyen (approved):
lgtm 🚀

app/types/profile.ts
```diff
@@ -1,19 +1,19 @@
...
-	hobbies?: string;
-	avoid?: string;
-	practical_whimsical?: number;
-	cozy_adventurous?: number;
-	minimal_luxurious?: number;
-	created_at: string;
-	updated_at: string;
-	onboarding_complete: boolean;
+  id: string;
+  display_name: string;
```
thomasnguyen, line 3: oh good catch haha

app/profile/LoadingCard.tsx
```diff
@@ -0,0 +1,25 @@
+import { Skeleton } from "@/components/Skeleton/Skeleton";
+
+const LoadingCard = () => {
+  return (
+    <article className="text-white px-4">
+      <div className="flex items-center rounded-t-2xl py-5 px-9 bg-groupCardGreen">
+        <Skeleton className="w-full h-10 rounded-full" />
+      </div>
+      <div className="bg-[#12433A] rounded-b-2xl grid grid-cols-1 lg:grid-cols-[minmax(550px,2fr)_minmax(300px,1fr)] gap-12 pl-9 pr-28 py-7">
```
alexappleget, line 9: something i noticed is that `xl:` instead of `lg:` works for most computer screens, but then we would also have to add css for the default (all other screen sizes) so that this adapts to all screens.

Ex: `xl:grid-cols-[minmax(550px,2fr)_minmax(300px,1fr)]`
and then default is just `grid-cols-...` whatever we want it to look like on mobile

alexappleget: this is nit ^

components/ProfileCard/ProfileCard.tsx
```diff
@@ -1,24 +1,80 @@
...
+      <div className="flex flex-col gap-2 text-xs">
+        <div className="flex items-center justify-between">
+          <div className={preference_value <= 50 ? "font-bold" : ""}>
+            {preferenceLeft}
+          </div>
+          <div className={preference_value >= 50 ? "font-bold" : ""}>
+            {preferenceRight}
+          </div>
+        </div>
+        <Progress indicatorClassName="bg-[#E8577D]" value={preference_value} />
```
alexappleget, line 59: i did not know you could do `bg-[#E8577D]` in tailwind. I have been using the config lmao

components/ProfileCard/ProfileCard.tsx
```diff
@@ -1,24 +1,80 @@
...
+
   return (
     <article className="text-white px-4">
       <div className="flex items-center rounded-t-2xl justify-between py-5 px-9 bg-groupCardGreen">
         <div className="flex items-center gap-4">
           <Avatar userAvatar="https://static.vecteezy.com/system/resources/previews/024/183/525/non_2x/avatar-of-a-man-portrait-of-a-young-guy-illustration-of-male-character-in-modern-color-style-vector.jpg" />
           <div>
-            <h1 className="text-lg font-bold">John Doe</h1>
+            <h1 className="text-lg font-bold">
+              {formatDisplayName(profile?.display_name || "")}
```
alexappleget, line 71: should we at least add something into the empty string so we know that this is working but the user doesnt have a name set. Like `"No Name Provided"`

app/profile/LoadingCard.tsx
jennifertieu: I'm open to changing it! I went with the mentality that as soon as the screen becomes larger (aka not table or mobile) we set it to the desktop view so that's why I went with `lg` since that would mean the screen was greater than 1024px.

We can go with `xl` instead for greater than 1280px but this is usually where I would ask the designer what they want @thomasnguyen 

https://tailwindcss.com/docs/responsive-design

components/ProfileCard/ProfileCard.tsx
jennifertieu: lol yes, it's an arbitrary value. Super useful for one-off colors: https://tailwindcss.com/docs/adding-custom-styles#using-arbitrary-values

I think its technically better to use the config though but I knew we were all putting colors in the config and so to prevent us having to resolve conflicts with the colors, I figured we could clean it up later

app/profile/LoadingCard.tsx
alexappleget: oh ok thats good to know! :)

components/ProfileCard/ProfileCard.tsx
jennifertieu: Good question....I think we definitely need to indicate something isn't right. I can add that for now!

jennifertieu: Updated

PR #60, opened 2024-12-20
Fix Add member to gift exchange

thomasnguyen:
This PR fixes the API route so that we can pass in a user id to add a certain user to a gift exchange.

You can use it like below
```
fetch('/api/gift-exchanges/8e3f26aa-59c6-49a1-bc76-91c956f59c01/members', {
  method: 'POST',
  headers: {
    'Content-Type': 'application/json'
  },
  body: JSON.stringify({
    user_id: '5f2a948d-b860-4de0-be9a-42eb0856a13a'
  })
})
.then(response => response.json())
.then(console.log)
.catch(console.error);
```

Screenshot 2024-12-19 at 10 32 41 PM

### DIFF
--- a/app/api/gift-exchanges/[id]/members/route.ts
+++ b/app/api/gift-exchanges/[id]/members/route.ts
@@ -60,10 +60,12 @@ export async function GET(
   }
 }
 
-// add a member to a gift exchange
-export async function POST(req: Request) {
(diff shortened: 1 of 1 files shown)


kepsteen (approved):
LGTM

PR #76, opened 2024-12-23
Alex/Gift-Suggestion-Feedback-Frontend

### LINKED ISSUE #50: [GiftSuggestion] Gift Suggestion Feedback Frontend
This ticket implements the GiftSuggestion feedback mechanism on the frontend

- [ ] When the user clicks on "not this", it should show the feedback options
- [ ] Users should go back or cancel if they want to
- [ ] When they select an option, it should fire the API event from the BE
- [ ] It should show loading skeletons while the API is waiting for a response.
- [ ] On response, show the new data and take out of "feedback mode"

Image

Image

alexappleget:
Closes #50 

https://github.com/user-attachments/assets/9c707d9a-8d05-4273-8ae7-c77520474b6f

### DIFF
--- a/components/GiftSuggestionCard/GiftSuggestionCard.tsx
+++ b/components/GiftSuggestionCard/GiftSuggestionCard.tsx
@@ -12,14 +12,14 @@ const GiftSuggestionCard: React.FC<GiftSuggestionCardProps> = ({ gift }) => {
       return (
         <FeedbackView handleFeedback={() => setIsShowingFeedback(false)} />
       );
-    } else {
-      return (
-        <GiftDetailsView
-          gift={gift}
-          handleFeedback={() => setIsShowingFeedback(true)}
-        />
-      );
     }
+
+    return (
+      <GiftDetailsView
+        gift={gift}
+        handleFeedback={() => setIsShowingFeedback(true)}
+      />
+    );
   };
   return (
     <Card className="bg-giftSuggestionsCardBackground h-96 w-80 flex flex-col justify-between m-5">
--- a/components/GiftSuggestionCard/GiftSuggestionCard.tsx
+++ b/components/GiftSuggestionCard/GiftSuggestionCard.tsx
@@ -1,90 +1,29 @@
-import { SquareArrowOutUpRight, ThumbsDown } from "lucide-react";
-import { Button } from "../Button/button";
-import {
-  Card,
-  CardContent,
-  CardDescription,
-  CardFooter,
-  CardHeader,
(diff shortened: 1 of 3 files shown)


components/GiftSuggestionCard/GiftSuggestionCard.tsx
```diff
@@ -8,48 +8,83 @@ import {
...
-      <CardFooter className="flex flex-col p-4">
-        <div className="flex justify-between w-full">
-          <Button className="text-sm w-32 h-9 bg-primaryButtonYelow70 hover:bg-primaryButtonYellow">
-            <SquareArrowOutUpRight /> View
-          </Button>
-          <Button className="text-sm w-32 h-9 text-giftSuggestionDarkGreen bg-gray-100 hover:bg-gray-200">
-            <ThumbsDown />
-            Not This
-          </Button>
+      {feedback ? (
```
thomasnguyen, line 19: nit: isShowingFeedback or isFeedbackView would be a better name for this since its a boolean variable

thomasnguyen, line 19: personally woudn't use a ternary operator here with all of this jsx. it makes it harder to read

thomasnguyen (approved):
have some comments but considering we are so close to the hackathon deadline, not going to block it.

LGTM

components/GiftSuggestionCard/GiftSuggestionCard.tsx
alexappleget: It is a ternary operator

alexappleget: I think so too. Will change when I get the chance

thomasnguyen: yeah, i said, i wouldn't use it here

alexappleget: OH hahaha my brain saw "would"

HoldUpFjord: In the feature-video, it doesn't showcase these parts of the ticket criteria: 
 [ ] It should show loading skeletons while the API is waiting for a response.
 [ ] On response, show the new data and take out of "feedback mode"
 
 Just want to double check this was either implemented or purposely bypassed

alexappleget: theres another ticket for the backend. i was going to do it then

PR #97, opened 2024-12-29
alex/fix-warning-modal-flicker

### LINKED ISSUE #96: Bug: Fix WarningModal flicker when members navigate to their groups
When a user is already a member of a group and the group is 'pending' their screen will flicker the `WarningModal` component for a brief second and disappear.

alexappleget:
Closes #96 

Reasoning for this:
- when a user is already a member of a gift exchange group, the `WarningModal` component will flicker for a brief second before disappearing.

I changed the default useState from `false` to `null` and then changed `!isUserAMember` to `isUserAMember === false` because is the default is `false` then when the page loads it will show it regardless until it goes through the apiFetching and set it to `true` to cause it to disappear.

Setting the default value to `null` will tell it not to display anything until it knows for sure that the person isn't a member of the group and then itll show once the page loads.


kepsteen (approved):
LGTM

HoldUpFjord (approved):
Looks like there's some conflicts that need to be resolved at `app/gift-exchanges/[id]/page.tsx`

PR #103, opened 2024-12-30
add images

thomasnguyen:
- on gift generation, pull images with google search API
- get first search result, we are assuming its the best result
- store in db blob
- update gift suggestion component to expect image with fallback

https://github.com/user-attachments/assets/11a08f81-3fb7-42c4-8a05-392047cf78f4

Fallback if image doesnt exist

Screenshot 2024-12-29 at 8 05 56 PM

### DIFF
--- a/lib/generateAndStoreSuggestion.ts
+++ b/lib/generateAndStoreSuggestion.ts
@@ -1,5 +1,6 @@
 import { SupabaseClient } from "@supabase/supabase-js";
 import { openai } from "../app/api/openaiConfig/config";
+import { getAmazonImage } from "./getAmazonImage";
 
 export async function generateAndStoreSuggestions(
   supabase: SupabaseClient,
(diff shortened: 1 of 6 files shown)


lib/generateAndStoreSuggestion.ts
```diff
@@ -52,13 +53,34 @@ export async function generateAndStoreSuggestions(
...
+      if (startIndex !== -1) {
+        jsonContent = jsonContent.slice(startIndex);
+      }
+    }
+    if (!jsonContent.endsWith(']')) {
+      const endIndex = jsonContent.lastIndexOf(']');
+      if (endIndex !== -1) {
+        jsonContent = jsonContent.slice(0, endIndex + 1);
+      }
+    }
```
thomasnguyen, line 76: parsing the response can sometimes be annyoing. hacked this together. but probably not needed most of the time

alexappleget (approved):
LGTM

PR #111, opened 2025-01-03
Dominick/gift exchange member button

### LINKED ISSUE #43: [GiftExchange] Gift Exchange "Draw members" button logic
- [x] Add button to "draw members now" now to the Gift Exchange header card
- [x] It should show only if the GiftExchange hasn't been drawn yet
- [x] It should call the draw GiftExchange API call done in #42  
- [x] It should refresh the page with new data

HoldUpFjord:
Closes #43 

https://github.com/user-attachments/assets/4ba34e38-aed2-41a6-a373-a99deb9e6c53

- [x] Add button to "draw members now" now to the Gift Exchange header card
- [x] It should show only if the GiftExchange hasn't been drawn yet
- [x] It should call the draw GiftExchange API call done in
https://github.com/guardiansofthegit-collab/elecretanta/issues/42
- [x] It should refresh the page with new data
- [x] User cannot start a draw without 3 or more users in group.


app/gift-exchanges/[id]/GiftExchangeHeader.tsx
```diff
@@ -1,133 +1,224 @@
...
-			case "completed":
-				return "Completed";
-			case "cancelled":
-				return "Cancelled";
-			default:
-				return "Open";
-		}
-	};
+      const data = await response.json();
+      location.reload();
```
Comment (HoldUpFjord) [inline app/gift-exchanges/[id]/GiftExchangeHeader.tsx:118]: I don't _love_ using this to hard refresh the page. Ideally we would update the page state with a use-effect of some sort. It works, but there are some other alternatives if time allows discussed in this stack overflow page: https://stackoverflow.com/questions/75655010/router-refresh-not-refreshing-in-next-13. 

I went with the minimal effort approach due to time.

Comment (thomasnguyen) [reply app/gift-exchanges/[id]/GiftExchangeHeader.tsx:118]: Yeah ideally we would have some refetching logic but I dont think its worth implementing right now (if we used react-query this would be a piece of cake).

Reloading like this is perfectly fine with me!

thomasnguyen (approved):

app/gift-exchanges/[id]/GiftExchangeHeader.tsx
Comment (HoldUpFjord) [reply app/gift-exchanges/[id]/GiftExchangeHeader.tsx:118]: Sounds good thanks dude! I'll go ahead and merge it in then. 🙏

PR #301, opened 2025-03-24
fix: upgraded nextjs version

alexappleget:
Updated the nextjs version to handle the security vulnerability

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
     "date-fns": "^3.6.0",
     "lucide-react": "^0.464.0",
     "motion": "^11.15.0",
-    "next": "15.0.3",
+    "next": "^15.2.3",
     "openai": "^4.73.1",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -30,7 +30,7 @@
     "date-fns": "^3.6.0",
     "lucide-react": "^0.464.0",
     "motion": "^11.15.0",
-    "next": "15.0.3",
+    "next": "^15.2.3",
```
shashilo, line 33: We should not auto upgrade minor and patches. Remove the caret. We'll handle these upgrades when we need to.

shashilo (changes requested):

package.json
alexappleget: fixed

kepsteen (approved):
LGTM 🚀

shashilo (approved):

PR #305, opened 2025-04-09
Feat: added prettier formatting as well as pnpm to the project

### LINKED ISSUE #304: Add Prettier to the codebase
Set up Prettier formatting for the codebase as well as pnpm

alexappleget:
Closes #304 

We needed to setup prettier formatting as well as pnpm for the project. I set up the config file, as well as, the ignore file for prettier. And then did pnpm i to set up the pnpm-lock.yaml file.

Note: when pushing the changes there was a build error with the framer motion dependency. So, I had to do `pnpm add framer-motion` to get it added to the `pnpm-lock` file and the `package.json` file.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,0 +1,5 @@
+.github/
+*.config.js
+.prettierrc.js
+*.json
+*.md
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,5 +1,8 @@
 .github/
 *.config.js
 .prettierrc.js
-*.json
 *.md
+settings.json
+package.json
+pnpm-lock.yaml
+.test-results/
--- a/package.json
+++ b/package.json
@@ -28,6 +28,7 @@
     "cmdk": "^1.0.0",
     "cross-env": "^7.0.3",
     "date-fns": "^3.6.0",
+    "framer-motion": "^12.6.3",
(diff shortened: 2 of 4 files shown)


.prettierignore
```diff
@@ -0,0 +1,5 @@
+.github/
+*.config.js
+.prettierrc.js
+*.json
```
mhchen, line 4: Why ignore these? Seems worth having them.

package.json
```diff
@@ -28,6 +28,7 @@
     "cmdk": "^1.0.0",
     "cross-env": "^7.0.3",
     "date-fns": "^3.6.0",
+    "framer-motion": "^12.6.3",
```
mhchen, line 31: Is this intentional in this PR?

mhchen (approved):

.prettierignore
alexappleget: I was following the documentation we had set up for GiS: https://docs.gridironsurvivor.com/?path=/docs/technical-planning-documents-prettier--docs

I believe it's so that prettier doesn't affect the formatting of those certain files

alexappleget:
@mhchen I made changes to both prettier files to mimic what GiS prettier files had.

shashilo (approved):

PR #311, opened 2025-04-11
fix: address eslint errors in dialog component

hdkassamali:
Closes #255 

Addresses the following eslint errors in the dialog Component:

- Line 1:1
  - Error: missing header header/header
- Line 56:22
  - Error: Missing JSDoc comment. jsdoc/require-jsdoc
- LIne 59:42
  - Error: Missing return type on function. @typescript-eslint/explicit-function-return-type
- Line 70:22
  - Error: Missing JSDoc comment. jsdoc/require-jsdoc
- Line 73:42
  - Error: Missing return type on function. @typescript-eslint/explicit-function-return-type

**NOTE:** The issue mentioned these other errors, but they may have been addressed already. When I ran `npx eslint ./components/Dialogue/dialog.tsx`, these errors did not show up in the terminal:
  
- Line 20:6
  - Error: 'className' is missing in props validation react/prop-types
- Line 35:6
  - Error: 'className' is missing in props validation react/prop-types
- Line 87:6
  - Error: 'className' is missing in props validation react/prop-types
- Line 102:6
  - Error: 'className' is missing in props validation react/prop-types

PR #316, opened 2025-04-12
lillian/252 fix eslint header error for card component

### LINKED ISSUE #252: Address ESlint errors in the file: ./components/Card/card.tsx
## Replicating errors:
- Open terminal
- execute terminal command `npx eslint ./components/Card/card.tsx`

## Existing errors:
- Line 1:1 
  -  Error: missing header  header/header

lillian-nguyen:
Addresses following eslint error in ./components/Card/card.tsx : 

- Line 1:1 - Missing header header/header 

Additional changes: 
- Destructure type imports from React
- Add JSDoc comments

closes #252


kepsteen (approved):
LGTM

kepsteen (dismissed):
LGTM

components/Card/card.tsx
```diff
@@ -1,3 +1,6 @@
+// Copyright (c) Gridiron Survivor.
+// Licensed under the MIT License.
+
 import * as React from 'react';
 
 import { cn } from '@/lib/utils';
```
shashilo, line 6: JS Docs linter should've cause that this function needs to be documented. Can you check with @hdkassamali and @destocot how they got that setup.

shashilo (changes requested):
@lillian-nguyen Please review my comment.

components/Card/card.tsx
hdkassamali: Screenshot 2025-04-14 at 3 10 26 PM

@shashilo @lillian-nguyen, I also only saw that header error.

hdkassamali: #311 
I'm not entirely sure, but it seems like I had similar functions in the dialog component, where I didn't have to document my functions. 

For example: `Dialog Overlay` and `DialogContent` didn't cause any lint errors, but `DialogHeader` and `DialogFooter` did. The only difference I see between the header and footer functions is that I'm passing in `className and ...props` as parameters before I pass in the elements, instead of after. 

It seems like all of the functions in the `card.tsx` file pass in className and ...props after the elements, and maybe that's why there are no errors

kepsteen (approved):
LGTM

shashilo (approved):

PR #325, opened 2025-04-14
Alex/set-up-playwright

### LINKED ISSUE #324: Set up playwright

alexappleget:
Closes #324 

1. Installed playwright
2. Deleted example test folders and files
3. Deleted duplicate `node_modules/` in the .gitignore file as it's already in there


.gitignore
```diff
@@ -39,3 +39,9 @@ yarn-error.log*
 # typescript
 *.tsbuildinfo
 next-env.d.ts
+
+# Playwright
+/test-results/
```
shashilo, line 44: Why are we ignoring `test-results`, `playwrigt-report` and `blob-report`?

shashilo (changes requested):

.gitignore
alexappleget: It is what got auto put here when installing playwright and I wanted to match what gridiron had

PR #338, opened 2025-04-16
khurram/create-unit-test-for-badge

### LINKED ISSUE #178: Create unit test for badge.tsx

destocot:
Closes #178

### Before:
The Badge component (`components/Badge/badge.tsx`) was missing unit tests to verify functionality:
- No tests for rendering different badge variants
- No tests for default variant behavior
- No tests for applying custom class names

### After:
- extracted `variants` object from Button to be reusable in unit tests.

Created unit tests in `components/Badge/badge.test.tsx` that verify:
- Badge renders with default variant when none is provided
- Each badge variant renders correctly with proper styling
- Custom classNames are applied alongside variant styles
- Test for custom attributes

image

### Summary:
This PR adds unit test coverage for the Badge component


components/Badge/badge.test.tsx
```diff
@@ -0,0 +1,46 @@
...
+] as NonNullable<BadgeProps['variant']>[];
+
+describe('Badge', () => {
+  for (let i = 0; i < variantTestCases.length; i++) {
+    const variant = variantTestCases[i];
+
+    it(`renders variant: ${variant}`, () => {
+      render(<Badge variant={variant}>{variant}</Badge>);
+      const badge = screen.getByText(variant);
+      expect(badge).toBeInTheDocument();
```
alexappleget, line 21: this is only testing that a specific text shows up on screen, but it's not actually testing that the Badge shows up.

You want to add a test id to the actual rendered component inside the test, grab that directly, and make sure it's in the document. Then you can test the variant.

components/Badge/badge.test.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+      const classNames = badgeVariants({ variant });
+      expect(badge).toHaveClass(classNames);
```
alexappleget, line 24: as per my comment above, once you have the specific test id grabbed then you can accurately test that the component has what it needs.

components/Badge/badge.test.tsx
```diff
@@ -0,0 +1,45 @@
+// Copyright (c) Gridiron Survivor.
+// Licensed under the MIT License.
+
+import { Badge, BadgeProps, badgeVariants } from './badge';
+import { render, screen } from '@testing-library/react';
+
+const variantTestCases = [
```
shashilo, line 7: Why are we creating these variant cases when they already exist in the main component?

components/Badge/badge.test.tsx
```diff
@@ -0,0 +1,45 @@
...
+      render(<Badge variant={variant}>{variant}</Badge>);
+      const badge = screen.getByTestId('badge');
+      expect(badge).toBeInTheDocument();
+
+      const classNames = badgeVariants({ variant });
+      expect(badge).toHaveClass(classNames);
+    });
+  }
+
+  it('renders with default variant when none is provided', () => {
```
shashilo, line 28: For default renders, what are the required props? We should be checking that each of these props are expected to have the render or value we're looking for.

components/Badge/badge.test.tsx
```diff
@@ -0,0 +1,45 @@
...
+  });
+
+  it('applies custom className alongside variant styles', () => {
+    render(<Badge className="custom-class">custom</Badge>);
+    const badge = screen.getByTestId('badge');
+    const classNames = badgeVariants({ variant: 'default' });
+
+    expect(badge).toHaveClass(classNames);
+    expect(badge).toHaveClass('custom-class');
+  });
```
shashilo, line 44: We need a separate test to check for the child props we pass in to render with the component.

shashilo (changes requested):

components/Badge/badge.test.tsx
destocot: the badge component has no required props

components/Badge/badge.test.tsx
destocot: there is no way to retrieve all the variant names from the component itself so I hardcoded an array that is relatively type safe to support these tests.

unless, i am misunderstanding what you mean here

components/Badge/badge.test.tsx
destocot: separate test will be made to specifically test for children prop

components/Badge/badge.test.tsx
destocot: these variants are not accsssible by what is currently in the component as far as i know

Danielle254 (dismissed):
Much improved test descriptions, well done! I'm going to let @shashilo address the variant cases in his review.

components/Badge/badge.test.tsx
shashilo: They should be accessible. Research how to get the variants from a ShadCN component.

components/Badge/badge.test.tsx
```diff
@@ -0,0 +1,61 @@
...
+describe('Badge', () => {
+  it('renders the component in the document', () => {
+    render(<Badge />);
+
+    const badge = screen.getByTestId('badge');
+
+    expect(badge).toBeInTheDocument();
+  });
+
+  it('renders with default variant when none is provided', () => {
```
shashilo, line 23: This is the same exact test as the default one. Combine them.

shashilo (changes requested):

kepsteen (approved):

components/Badge/badge.test.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+    const badge = screen.getByTestId('badge');
+
+    expect(badge).toBeInTheDocument();
+
+    const classNames = badgeVariants({ variant: 'default' });
+    expect(badge).toHaveClass(classNames);
+  });
+
+  for (let i = 0; i < variantNames.length; i++) {
```
shashilo, line 24: Separate this test to describe each variant. When this test fails, it should only fail if a variant's attributes are incorrect. Right now, it's part of the default test.

components/Badge/badge.tsx
```diff
@@ -38,7 +40,11 @@ export interface BadgeProps
  */
 const Badge = ({ className, variant, ...props }: BadgeProps): JSX.Element => {
   return (
-    <div className={cn(badgeVariants({ variant }), className)} {...props} />
+    <div
+      className={cn(badgeVariants({ variant }), className)}
+      data-testid="badge"
+      {...props}
```
shashilo, line 46: Need a test for custom attributes as well.

shashilo (changes requested):

components/Badge/badge.tsx
destocot: Can you clarify what you mean by custom attributes here?

components/Badge/badge.test.tsx
destocot: It is it's own set of tests the for loop is outside the `it('renders the component with default variant', () => {})` block

Danielle254 (approved):

components/Badge/badge.test.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+const variantTestCases = [
+  'default',
+  'destructive',
+  'outline',
+  'secondary',
+] as NonNullable<BadgeProps['variant']>[];
+
+describe('Badge', () => {
+  it('renders without crashing', () => {
```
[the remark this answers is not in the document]
destocot: fixed

components/Badge/badge.test.tsx
destocot: fixed

shashilo (approved):

alexappleget (approved):

PR #345, opened 2025-04-16
Hector/175 Created unit tests for AvatarBody

### LINKED ISSUE #175: Create unit test for AvatarBody.tsx

suzuhe481:
Closes #175 

Creating initial unit test for AvatarBody.

Run test with
```pnpm run test ./components/Avatar/AvatarBody/AvatarBody.test.tsx```

## Tests
- Tests that component renders with a test child.
- Tests that component renders without children.

### CODE REVIEW

File: components/Avatar/AvatarBody.test.tsx [hunk 1]
```diff
@@ -0,0 +1,25 @@
...
+describe('AvatarBody Component', () => {
+  it('should render correctly', () => {
+    render(<AvatarBody />);
+
+    const avatarBodyElement = screen.getByTestId('avatar-body');
+
+    expect(avatarBodyElement).toBeInTheDocument();
+  });
+
+  it('should render correctly with passed children', () => {
```
alexappleget, line 16: You don't even really need this test. You already tested that it renders correctly

alexappleget (changes requested):

suzuhe481 (commented):
Removed the second test.

components/Avatar/AvatarBody/AvatarBody.test.tsx
```diff
@@ -0,0 +1,29 @@
+// Copyright (c) Gridiron Survivor.
+// Licensed under the MIT License.
+
+import { AvatarBody } from './AvatarBody';
+import { render, screen } from '@testing-library/react';
+
+describe('AvatarBody Component', () => {
```
kepsteen, line 7: We also need to test that any additional props or classes that we pass down to this component are being rendered properly.

`<AvatarBody custom-prop className="custom-class" />`

kepsteen (dismissed):
🚀

components/Avatar/AvatarBody/AvatarBody.tsx
```diff
@@ -14,6 +17,7 @@ const AvatarBody = React.forwardRef<
       'relative flex h-10 w-10 shrink-0 overflow-hidden rounded-full',
       className,
     )}
+    data-testid="avatar-body"
     {...props}
```
shashilo, line 21: Need 1 more test for custom attributes to be visible in the component when passed via prop.

shashilo (changes requested):

components/Avatar/AvatarBody/AvatarBody.tsx
suzuhe481: I'm commenting this here for reference on something I learned about attribute names.

I'm was looking into how custom attributes are passed into components since this is wrapped in an AvatarPrimitive. From what I saw, you can only pass custom props as long as it has a hyphen, because it follows HTML specification in how attributes are named using kebab-case, even though we're using react.

So an attribute like this won't work.
- customProp

But these will.
- custom-prop
- Custom-prop
- data-custom-prop

Since we have testid in this format ```data-testid```, I wrote the test in a similar way with the attribute ```data-custom-prop```, but it should still work with any kebab-case attribute.

shashilo (approved):

alexappleget (approved):

nickytonline (approved):
Thanks for fixing the merge conflicts! 🚢

PR #348, opened 2025-04-16
Hector/176 Created unit test for AvatarFallback

### LINKED ISSUE #176: Create unit test for AvatarFallback.tsx

suzuhe481:
Closes #176

Creating the unit test for AvatarFallback.

Run test with
```pnpm run test ./components/Avatar/AvatarFallback/AvatarFallback.test.tsx```

## Tests
- Tests that AvatarFallback renders correctly with passed children.
- Tests that AvatarFallback renders correctly with no children.

## Notes
- AvatarFallback is only used in the Avatar.tsx component where it gets passed an img element with a hardcoded src and alt attributes. **Would it be better to move that img element directly inside the AvatarFallback.tsx component?**

### CODE REVIEW

**Review by kepsteen — DISMISSED** [review]
LGTM

**Review by alexappleget — APPROVED** [review]
LGTM

File: components/Avatar/AvatarFallback/AvatarFallback.test.tsx [hunk 1]
```diff
@@ -0,0 +1,19 @@
+// Copyright (c) Gridiron Survivor.
+// Licensed under the MIT License.
+
+import Avatar from '../Avatar';
+import { render, screen } from '@testing-library/react';
+
+describe('AvatarFallback Component', () => {
+  it('should render correctly with an image', () => {
```
shashilo, line 8: If you look at the props that can be passed through a component, we need to test that those will change and update when we change them.

components/Avatar/AvatarFallback/AvatarFallback.test.tsx
```diff
@@ -0,0 +1,19 @@
...
+    // Avatarfallback must be used within Avatar
+    render(<Avatar userAvatar={''} />);
+
+    const avatarFallbackElement = screen.getByTestId('avatar-fallback');
+    const imgElement = screen.getByAltText("default avatar");
```
shashilo, line 13: Why are we testing for alt test in this component?

components/Avatar/AvatarFallback/AvatarFallback.test.tsx
```diff
@@ -0,0 +1,36 @@
...
+    );
+
+    const avatarFallbackElement = screen.queryByTestId('avatar-fallback');
+    const imgElement = screen.queryByRole('img');
+
+    expect(avatarFallbackElement).toBeInTheDocument();
+    expect(imgElement).toBeInTheDocument();
+  });
+
+  it('renders the component with no children', () => {
```
kepsteen, line 25: We also need to test that the props we pass down through this component are being added to the component, and that any additional classes we add are being passed down as well

kepsteen (changes requested):

components/Avatar/AvatarFallback/AvatarFallback.test.tsx
suzuhe481: When adding test props and classnames to the test, can I make it anything for the purpose of testing? Or should it be similar to what AvatarFallback is already using?

eg. Passing the classname as "flex h-full w-full items-center justify-center rounded-full bg-muted"

suzuhe481: Should it also be it's own test or can it be combined with the first?

kepsteen: its own test

components/Avatar/AvatarFallback/AvatarFallback.test.tsx
```diff
@@ -0,0 +1,50 @@
...
+  it('renders the component with image as child', () => {
+    render(
+      <AvatarPrimitiveRoot>
+        <AvatarFallback>
+          <img src="fallback-image.jpg" alt="default avatar" />
+        </AvatarFallback>
+      </AvatarPrimitiveRoot>,
+    );
+
+    const avatarFallbackElement = screen.queryByTestId('avatar-fallback');
```
kepsteen, line 18: Sorry, I missed this before, but let's use `getByTestId` instead of `queryByTestId`. The reason is that `getByTestId` will throw an error if the element is not found, whereas query will not. We want to use `queryByTestId` only when we want to assert that the element is not there.

components/Avatar/AvatarFallback/AvatarFallback.test.tsx
```diff
@@ -0,0 +1,50 @@
...
+  it('renders the component with custom classes', () => {
+    render(
+      <AvatarPrimitiveRoot>
+        <AvatarFallback className="custom-class">
+          <img src="fallback-image.jpg" alt="default avatar" />
+        </AvatarFallback>
+      </AvatarPrimitiveRoot>,
+    );
+
+    const avatarFallbackElement = screen.queryByTestId('avatar-fallback');
```
kepsteen, line 34: use `getByTestId` here as well

components/Avatar/AvatarFallback/AvatarFallback.test.tsx
```diff
@@ -0,0 +1,50 @@
...
+  });
+
+  it('renders the component with no children', () => {
+    render(
+      <AvatarPrimitiveRoot>
+        <AvatarFallback />
+      </AvatarPrimitiveRoot>,
+    );
+
+    const avatarFallbackElement = screen.queryByTestId('avatar-fallback');
```
kepsteen, line 46: and here

kepsteen (changes requested):

kepsteen (approved):
🚀

components/Avatar/Avatar.tsx
```diff
@@ -1,5 +1,5 @@
 import { AvatarImage } from './AvatarImage';
-import { AvatarFallback } from './AvatarFallback';
+import { AvatarFallback } from './AvatarFallback/AvatarFallback';
```
shashilo, line 2: Nice job decoupling this component.

components/Avatar/AvatarFallback/AvatarFallback.test.tsx
```diff
@@ -0,0 +1,50 @@
...
+    render(
+      <AvatarPrimitiveRoot>
+        <AvatarFallback>
+          <img src="fallback-image.jpg" alt="default avatar" />
+        </AvatarFallback>
+      </AvatarPrimitiveRoot>,
+    );
+
+    const avatarFallbackElement = screen.getByTestId('avatar-fallback');
+    const imgElement = screen.getByRole('img');
```
shashilo, line 19: Instead of getting it by role, setup a data-testid.

shashilo (changes requested):

components/Avatar/AvatarFallback/AvatarFallback.test.tsx
suzuhe481: I see 2 ways of doing this but which one do you prefer?

- I could just add the data-id to the img element which is in Avatar.tsx. 
- I could move the img element from Avatar.tsx to AvatarFallback.tsx. If I do this, I should also add src and alt attributes to AvatarFallback to pass them to the image, similar to how the AvatarImage component works. **(If I do it this way, should I use \ or \<Image\> from radix-ui?) The AvatarImage component uses \<Image\>.**

shashilo: Add a static data-testid to each element

components/Avatar/AvatarFallback/AvatarFallback.test.tsx
```diff
@@ -0,0 +1,54 @@
...
+// Licensed under the MIT License.
+
+import { AvatarFallback } from './AvatarFallback';
+import { Root as AvatarPrimitiveRoot } from '@radix-ui/react-avatar';
+import { render, screen } from '@testing-library/react';
+
+describe('AvatarFallback Component', () => {
+  it('renders the component with image as child', () => {
+    render(
+      <AvatarPrimitiveRoot>
```
shashilo, line 11: Why do we need AvatarPrimitiveRoot?

suzuhe481: The AvatarFallback component has to be inside Avatar or I'd get an error. I didn't want to directly use Avatar directly because I want to test Fallback without using another component. I used another PR as reference, to use the Root component of Avatar from Radix-UI instead of the project's Avatar.

nickytonline:
The tests look good. One thing to consider from an a11y perspective is to set the avatar fallback to have a `aria-hidden="true"` prop if there is no image, i.e. `children` prop. This will stop a screen reader from announcing the purely decorative avatar fallback div.

```diff
const AvatarFallback = forwardRef<
  ElementRef<typeof AvatarPrimitive.Fallback>,
  ComponentPropsWithoutRef<typeof AvatarPrimitive.Fallback>
- >(({ className, ...props }, ref) => (
+ >(({ className, children, ...props }, ref) => (
  <AvatarPrimitive.Fallback
    ref={ref}
    className={cn(
      'flex h-full w-full items-center justify-center rounded-full bg-muted',
      className,
    )}
+    aria-hidden={children ? undefined : true}
    data-testid="avatar-fallback"
-    {...props}
-  />
+    {...props}
+  >
+    {children}
+  </AvatarPrimitive.Fallback>
));
```

shashilo (approved):

alexappleget (approved):

PR #349, opened 2025-04-17
Julianna/257-ESLint-Errors-Form

mathematiCode:
Here are the errors and how I addressed each of them: 
Full disclosure I used Cursor a lot for this because typescript is still a pretty new to me and I haven't used react-hook-form or shadcn before. 

Line 1:1
Error: missing header header/header - I added the header 
Line 31:19
Error: Missing JSDoc comment. jsdoc/require-jsdoc
Line 35:1
Error: Expected indentation of 4 spaces but found 2. indent
Line 36:1
Error: Expected indentation of 2 spaces but found 0. indent
Line 36:42
Error: Missing return type on function. @typescript-eslint/explicit-function-return-type  - I added it on line 55, it's a JSX.element
Line 38:39
Error: The object passed as the value prop to the Context provider (at line 38) changes every render. To fix this consider wrapping it in a useMemo hook. react/jsx-no-constructed-context-values - I wrapped it in a useMemo hook defined in the value variable. 
Line 44:22
Error: Missing JSDoc comment. jsdoc/require-jsdoc - Added it 
Line 44:25
Error: Missing return type on function. @typescript-eslint/explicit-function-return-type - I added it on line 69, it's a FormFieldContextValue 
Line 82:38
Error: The object passed as the value prop to the Context provider (at line 82) changes every render. To fix this consider wrapping it in a useMemo hook. react/jsx-no-constructed-context-values - I wrapped it in a useMemo hook defined in the value variable. 
Line 92:6
Error: 'className' is missing in props validation react/prop-types - I created a type that includes className as an optional prop. I could change it to an interface though if anyone has a preference for one over the other.

### DIFF
--- a/components/Form/form.tsx
+++ b/components/Form/form.tsx
@@ -14,6 +14,7 @@ import {
   FieldValues,
   FormProvider,
   useFormContext,
+  FieldError,
 } from 'react-hook-form';
 
 import { cn } from '@/lib/utils';
@@ -26,25 +27,45 @@ type FormFieldContextValue<
   TName extends FieldPath<TFieldValues> = FieldPath<TFieldValues>,
 > = {
   name: TName;
+  id?: string;
+  formItemId?: string;
+  formDescriptionId?: string;
+  formMessageId?: string;
+  error?: FieldError;
+  isTouched?: boolean;
+  isDirty?: boolean;
 };
 
 const FormFieldContext = React.createContext<FormFieldContextValue>(
   {} as FormFieldContextValue,
 );
 
+/**
+ * A form field component that wraps react-hook-form's Controller.
+ * @template TFieldValues - Type for the form values object
+ * @template TName - Type for the field name, must be a valid path in TFieldValues
+ * @param {ControllerProps<TFieldValues, TName>} props - Props for the Controller component
+ * @returns {JSX.Element} A form field component
+ */
 const FormField = <
   TFieldValues extends FieldValues = FieldValues,
   TName extends FieldPath<TFieldValues> = FieldPath<TFieldValues>,
 >({
-  ...props
-}: ControllerProps<TFieldValues, TName>): JSX.Element => {
+    ...props
+  }: ControllerProps<TFieldValues, TName>): JSX.Element => {
+  const value = React.useMemo(() => ({ name: props.name }), [props.name]);
+  
   return (
-    <FormFieldContext.Provider value={{ name: props.name }}>
+    <FormFieldContext.Provider value={value}>
       <Controller {...props} />
     </FormFieldContext.Provider>
   );
 };
 
+/**
+ * Custom hook that provides form field context and state information.
+ * @returns {FormFieldContextValue} Form field information object
+ */
 const useFormField = (): FormFieldContextValue => {
   const fieldContext = React.useContext(FormFieldContext);
   const itemContext = React.useContext(FormItemContext);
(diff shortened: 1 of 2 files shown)


components/Form/form.tsx
```diff
@@ -22,28 +26,48 @@ type FormFieldContextValue<
...
+  formItemId?: string;
+  formDescriptionId?: string;
+  formMessageId?: string;
+  error?: FieldError;
+  isTouched?: boolean;
+  isDirty?: boolean;
 };
 
-const FormFieldContext = React.createContext<FormFieldContextValue>(
+const FormFieldContext = createContext<FormFieldContextValue>(
```
shashilo, line 38: Need to add JS Docs to each function.

components/Form/form.tsx
```diff
@@ -22,28 +26,48 @@ type FormFieldContextValue<
...
+ */
 const FormField = <
   TFieldValues extends FieldValues = FieldValues,
   TName extends FieldPath<TFieldValues> = FieldPath<TFieldValues>,
 >({
-  ...props
-}: ControllerProps<TFieldValues, TName>) => {
+    ...props
+  }: ControllerProps<TFieldValues, TName>): JSX.Element => {
+  const value = useMemo(() => ({ name: props.name }), [props.name]);
```
shashilo, line 55: Why are we adding this?

components/Form/form.tsx
```diff
@@ -68,27 +92,34 @@ type FormItemContextValue = {
...
 );
 
-const FormItem = React.forwardRef<
+const FormItem = forwardRef<
   HTMLDivElement,
   React.HTMLAttributes<HTMLDivElement>
 >(({ className, ...props }, ref) => {
-  const id = React.useId();
+  const id = useId();
+  const value = useMemo(() => ({ id }), [id]);
```
shashilo, line 104: Why are we adding this?

components/Form/form.tsx
```diff
@@ -68,27 +92,34 @@ type FormItemContextValue = {
...
-    <FormItemContext.Provider value={{ id }}>
+    <FormItemContext.Provider value={value}>
       <div ref={ref} className={cn('space-y-2', className)} {...props} />
     </FormItemContext.Provider>
   );
 });
 FormItem.displayName = 'FormItem';
 
-const FormLabel = React.forwardRef<
+type FormLabelProps = React.ComponentPropsWithoutRef<
```
shashilo, line 114: Why was this updated to use this type?

shashilo (changes requested):

components/Form/form.tsx
mathematiCode: @shashilo It fixes this error:
_The object passed as the value prop to the Context provider (at line 38) changes every render. To fix this consider wrapping it in a useMemo hook. react/jsx-no-constructed-context-values_ 
The props object looks like this: 
image
Each name value is set wherever the FormField component is called and doesn't change based on user input. It's just the name of the FormField. But I think some of the other key-value pairs in props could change more frequently. So we're using useMemo to avoid resetting the FormFieldContext each time any of the props change.

components/Form/form.tsx
mathematiCode: Same as above, it fixes this error:
_The object passed as the value prop to the Context provider (at line 82) changes every render. To fix this consider wrapping it in a useMemo hook. react/jsx-no-constructed-context-values_

components/Form/form.tsx
mathematiCode: I used this to add the className to the type. It wasn't using an interface before, just a regular type. Do you want me to make it an interface instead?

shashilo (approved):

kepsteen (approved):
LGTM. Nice work 👍

PR #351, opened 2025-04-18
khurram/create-unit-test-for-avatar-image

### LINKED ISSUE #177: Create unit test for AvatarImage.tsx

destocot:
Closes #177 

### Before:
The `AvatarImage` component (`components/Avatar/AvatarImage.tsx`) lacked unit tests:
- No tests for rendering component with `src` and `alt` properties
- No tests for applying custom class names

### After:
Created `components/Avatar/AvatarImage.test.tsx` to verify:
- The component renders the image element with correct `src` and `alt`
- Custom class names are applied properly
- The test passes reliably using a mock of `@radix-ui/react-avatar`’s `Image` component

### Additonal:
Radix’s `Avatar.Image` only renders its `` after the image load event. To work around this, the test mocks it to always render a plain `` so we can reliably validate behavior.

`AvatarImage `is also wrapped in `AvatarPrimativeRoot `to reflect real app usage, even though it's not required when the component is mocked.

image

### Summary:
This PR adds reliable unit test coverage for the `AvatarImage` component and handles Radix UI’s image rendering behavior through component mocking.


components/Avatar/AvatarImage.test.tsx
```diff
@@ -0,0 +1,39 @@
...
+import { Root as AvatarPrimativeRoot, Image } from '@radix-ui/react-avatar';
+import { render, screen } from '@testing-library/react';
+
+jest.mock('@radix-ui/react-avatar', () => ({
+  ...jest.requireActual('@radix-ui/react-avatar'),
+  Image: (props: typeof Image) => <img data-testid="avatar-image" {...props} />,
+}));
+
+describe('AvatarImage', () => {
+  it('renders with correct src and alt attributes', () => {
```
shashilo, line 11: The first test should always check the default state of the component. Then, it should test all the possible variants via number of props.

shashilo (changes requested):

components/Avatar/AvatarImage.tsx
```diff
@@ -18,11 +19,13 @@ const AvatarImage = forwardRef<
   ElementRef<typeof Image>,
   ComponentPropsWithoutRef<typeof Image>
 >(
-  ({ alt, className, ...props }, ref): JSX.Element => (
+  ({ alt, src, className, ...props }, ref): JSX.Element => (
     <Image
       ref={ref}
+      src={src}
```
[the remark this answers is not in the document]
destocot: I feel it is better to destructure this property from props, however, I will bring this up separately I will remove since it doesn't pertain directly to the ticket

components/Avatar/AvatarImage.test.tsx
```diff
@@ -0,0 +1,39 @@
+import { AvatarImage } from './AvatarImage';
+import { Root as AvatarPrimativeRoot, Image } from '@radix-ui/react-avatar';
+import { render, screen } from '@testing-library/react';
+
+jest.mock('@radix-ui/react-avatar', () => ({
```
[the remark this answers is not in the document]
destocot: I mentioned above in the PR:

> Radix’s Avatar.Image only renders its  after the image load event. To work around this, the test mocks it to always render a plain  so we can reliably validate behavior.

AvatarImage is also wrapped in AvatarPrimativeRoot to reflect real app usage, even though it's not required when the component is mocked.

components/Avatar/AvatarImage.test.tsx
```diff
@@ -0,0 +1,48 @@
...
+  ...jest.requireActual('@radix-ui/react-avatar'),
+  Image: (props: typeof Image) => <img data-testid="avatar-image" {...props} />,
+}));
+
+describe('AvatarImage', () => {
+  it('renders without crashing', () => {
```
Danielle254, line 11: do we need `without crashing`?

components/Avatar/AvatarImage.test.tsx
```diff
@@ -0,0 +1,48 @@
...
+    render(
+      <AvatarPrimativeRoot>
+        <AvatarImage />
+      </AvatarPrimativeRoot>,
+    );
+    const avatarImage = screen.getByTestId('avatar-image');
```
Danielle254, line 17: I don't see a `testId` added to the component in AvatarImage.tsx

kepsteen (approved):
LGTM

Danielle254 (dismissed):

components/Avatar/AvatarImage.test.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+  it('applies custom className along with default styles', () => {
+    render(
+      <AvatarPrimativeRoot>
+        <AvatarImage className="custom-class" />
+      </AvatarPrimativeRoot>,
+    );
+
+    const avatarImage = screen.getByTestId('avatar-image');
+    expect(avatarImage).toHaveClass('aspect-square h-full w-full');
```
shashilo, line 42: No reason to check for this. It's there by default. If the component renders and the data-testid exist, these classes will come along with it by default.

shashilo (changes requested):

components/Avatar/AvatarImage.test.tsx
destocot: My intent was to show that the classNames would be appended to and not overwritten I have removed the expect default classNames line per your feedback

kepsteen (approved):

components/Avatar/AvatarImage.test.tsx
Danielle254: @destocot please make this change

Danielle254 (changes requested):

components/Avatar/AvatarImage.test.tsx
destocot: fixed

components/Avatar/AvatarImage.test.tsx
shashilo: No need because the classes do not change. If the data-testify exist, the classes should as well.

shashilo (approved):

alexappleget (approved):

PR #366, opened 2025-04-24
Cody/184 dialog unit test

kepsteen:
- added unit tests for dialog component

Issue: Failing test to open the dialog with the keyboard

### DIFF
--- a/components/Dialogue/Dialog.test.tsx
+++ b/components/Dialogue/Dialog.test.tsx
@@ -1,4 +1,4 @@
-import { fireEvent, render, screen } from '@testing-library/react';
+import { fireEvent, render, screen, act } from '@testing-library/react';
 import {
   Dialog,
   DialogTrigger,
@@ -8,13 +8,16 @@ import {
   DialogDescription,
   DialogClose,
 } from './dialog';
-import React from 'react';
 
 describe('Dialog', () => {
   it('should render correctly with content when open', () => {
     render(
       <Dialog open={true}>
-        <DialogContent data-testid="dialog-content">Content</DialogContent>
+        <DialogContent data-testid="dialog-content">
+          <DialogTitle>Title</DialogTitle>
+          Content
+          <DialogDescription>Description</DialogDescription>
+        </DialogContent>
       </Dialog>,
     );
 
@@ -24,7 +27,11 @@ describe('Dialog', () => {
   it('should not render the content when closed', () => {
     render(
       <Dialog open={false}>
-        <DialogContent data-testid="dialog-content">Content</DialogContent>
+        <DialogContent data-testid="dialog-content">
+          <DialogTitle>Title</DialogTitle>
+          Content
+          <DialogDescription>Description</DialogDescription>
+        </DialogContent>
       </Dialog>,
     );
 
@@ -33,18 +40,56 @@ describe('Dialog', () => {
 
   it('should close the dialog if the Escape key is pressed', () => {
     render(
-      <Dialog open={true}>
-        <DialogContent data-testid="dialog-content">Content</DialogContent>
+      <Dialog defaultOpen={true}>
+        <DialogContent data-testid="dialog-content">
+          <DialogTitle>Title</DialogTitle>
+          Content
+          <DialogDescription>Description</DialogDescription>
+        </DialogContent>
       </Dialog>,
     );
+
     expect(screen.getByTestId('dialog-content')).toBeInTheDocument();
 
-    fireEvent.keyDown(window, {
+    fireEvent.keyDown(screen.getByTestId('dialog-content'), {
       key: 'Escape',
       code: 'Escape',
     });
 
     expect(screen.queryByTestId('dialog-content')).toBeNull();
+  });
+
+  it('should open the dialog with the keyboard', () => {
+    render(
+      <Dialog defaultOpen={false}>
+        <DialogTrigger data-testid="dialog-trigger">Trigger</DialogTrigger>
+        <DialogContent data-testid="dialog-content">
+          <DialogTitle>Title</DialogTitle>
+          Content
+          <DialogDescription>Description</DialogDescription>
+        </DialogContent>
+      </Dialog>,
+    );
+
+    const trigger = screen.getByTestId('dialog-trigger');
+    trigger.focus();
+
+    act(() => {
+      fireEvent.keyDown(trigger, {
+        key: 'Space',
+        code: 'Space',
+      });
+
+      fireEvent.keyUp(trigger, {
+        key: 'Enter',
+        code: 'Enter',
+        keyCode: 13,
+        which: 13,
+        bubbles: true,
+      });
+    });
+
+    expect(screen.getByTestId('dialog-content')).toBeInTheDocument();
   });
 
   describe('DialogTrigger', () => {
@@ -62,7 +107,11 @@ describe('Dialog', () => {
       render(
         <Dialog>
           <DialogTrigger data-testid="dialog-trigger">Trigger</DialogTrigger>
-          <DialogContent data-testid="dialog-content">Content</DialogContent>
+          <DialogContent data-testid="dialog-content">
+            <DialogTitle>Title</DialogTitle>
+            Content
+            <DialogDescription>Description</DialogDescription>
+          </DialogContent>
         </Dialog>,
       );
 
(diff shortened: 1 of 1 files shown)


components/Dialogue/Dialog.test.tsx
```diff
@@ -0,0 +1,175 @@
...
+
+    fireEvent.keyDown(screen.getByTestId('dialog-content'), {
+      key: 'Escape',
+      code: 'Escape',
+    });
+
+    expect(screen.queryByTestId('dialog-content')).toBeNull();
+  });
+
+  it('should open the dialog with the keyboard', () => {
```
kepsteen, line 62: This test is failing, so I think there might be an issue with how I'm simulating the keyDown Event?

hdkassamali: I may be wrong in this, but is it because you're missing this?

```
const dialogContent = screen.getByTestId("dialog-content");
dialogContent.focus();
```

I think if you add that after line 50, it should work.

components/Dialogue/Dialog.test.tsx
```diff
@@ -0,0 +1,185 @@
...
+      <Dialog open={true}>
+        <DialogContent data-testid="dialog-content">
+          <DialogTitle>Title</DialogTitle>
+          Content
+          <DialogDescription>Description</DialogDescription>
+        </DialogContent>
+      </Dialog>,
+    );
+
+    expect(screen.getByTestId('dialog-content')).toBeInTheDocument();
```
Danielle254, line 24: this does not fulfill your test as described. This only confirms the element is present, not that it renders with content.

components/Dialogue/Dialog.test.tsx
```diff
@@ -0,0 +1,185 @@
...
+      <Dialog open={false}>
+        <DialogContent data-testid="dialog-content">
+          <DialogTitle>Title</DialogTitle>
+          Content
+          <DialogDescription>Description</DialogDescription>
+        </DialogContent>
+      </Dialog>,
+    );
+
+    expect(screen.queryByTestId('dialog-content')).toBeNull();
```
Danielle254, line 38: I would add an additional test to see if the text is in the document. (to not be)

components/Dialogue/Dialog.test.tsx
```diff
@@ -0,0 +1,185 @@
...
+  });
+
+  it('should close the dialog if the Escape key is pressed', () => {
```
Danielle254, line 41: nice test!!

components/Dialogue/Dialog.test.tsx
```diff
@@ -0,0 +1,185 @@
...
+    );
+
+    expect(screen.getByTestId('dialog-content')).toBeInTheDocument();
+
+    fireEvent.keyDown(screen.getByTestId('dialog-content'), {
+      key: 'Escape',
+      code: 'Escape',
+    });
+
+    expect(screen.queryByTestId('dialog-content')).toBeNull();
```
Danielle254, line 59: I'm not familiar with `toBeNull()`. Does this mean we're checking that it is not rendered? That it is not in the document?

Danielle254 (changes requested):

components/Dialogue/Dialog.test.tsx
kepsteen: `queryByTestId` returns null if the element is not found. This is used in place of `getByTestId` which will throw an error if the element is not found

components/Dialogue/Dialog.test.tsx
```diff
@@ -0,0 +1,185 @@
...
+  DialogTrigger,
+  DialogContent,
+  DialogHeader,
+  DialogTitle,
+  DialogDescription,
+  DialogClose,
+} from './dialog';
+import userEvent from '@testing-library/user-event';
+describe('Dialog', () => {
+  it('should render correctly with content when open', () => {
```
Danielle254, line 13: what does "render correctly" mean and how are you testing it here?

Danielle254 (changes requested):

components/Dialogue/dialog.tsx
```diff
@@ -3,7 +3,13 @@
 
 'use client';
 
-import { HTMLAttributes, forwardRef, ElementRef, ComponentPropsWithoutRef, JSX } from 'react';
+import {
```
shashilo, line 6: Just like the other grouped components, ShadCN creates, decouple all the components into their own files. Then, write a unit test for each component.

shashilo (changes requested):

components/Dialog/DialogHeader.test.tsx
```diff
@@ -0,0 +1,85 @@
...
+import { DialogTitle } from './DialogTitle';
+
+describe('DialogHeader', () => {
+  it('should render with default props', () => {
+    render(
+      <Dialog defaultOpen={true}>
+        <DialogContent aria-describedby={undefined}>
+          <DialogTitle>Dialog Title</DialogTitle>
+          <DialogHeader>
+            <div>Header Content</div>
```
hdkassamali, line 17: Do you need to pass a <div> here and check for rendering children, since you already have a separate 'it' block testing for rendering children below?

hdkassamali (approved):
Added a comment inside `DialogHeader.test.tsx` more so out of curiosity as to why you might do that. Otherwise, LGTM

hdkassamali (approved):
LGTM

components/Dialog/Dialog.test.tsx
```diff
@@ -0,0 +1,161 @@
...
+      await user.click(screen.getByTestId('dialog-trigger'));
+      expect(screen.getByTestId('dialog-content')).toBeInTheDocument();
+
+      await user.click(screen.getByTestId('dialog-overlay'));
+      await waitFor(() => {
+        expect(screen.queryByTestId('dialog-content')).not.toBeInTheDocument();
+      });
+    });
+
+    test('should open when DialogTrigger is clicked and close by clicking DialogClose button', async () => {
```
Danielle254, line 114: why are you using `test` here? I thought our GIS convention was to always use `it`

Danielle254 (approved):

components/Dialog/Dialog.test.tsx
```diff
@@ -0,0 +1,161 @@
...
+  defaultOpen,
+  controlled = false,
+}: TestDialogComponentsProps) => {
+  const dialogProps = controlled
+    ? { open, onOpenChange }
+    : { defaultOpen, onOpenChange };
+
+  return (
+    <Dialog {...dialogProps}>
+      <DialogTrigger asChild data-testid="dialog-trigger">
```
shashilo, line 43: Why are we setting up the data-testid's in the test file?

components/Dialog/Dialog.test.tsx
```diff
@@ -0,0 +1,161 @@
...
+ *
+ * @param {Object} props - Component props
+ * @param {(open: boolean) => void} [props.onOpenChange] - Callback fired when dialog open state changes
+ * @param {boolean} [props.open] - Controls the open state (for controlled mode)
+ * @param {boolean} [props.defaultOpen] - Sets initial open state (for uncontrolled mode)
+ * @param {boolean} [props.controlled=false] - Whether to use controlled or uncontrolled mode
+ * @returns {JSX.Element} The test dialog component
+ */
+
+const TestDialogComponents = ({
```
shashilo, line 31: I would setup the component in the before each instead of a global variable. If there was a prop that changed, it would be cached. Opposed to restarting from a new instance for every test.

shashilo (changes requested):

components/Dialog/Dialog.test.tsx
kepsteen: This component is imported directly from the Radix Library, is there a way to add the `data-testid` to it?
```tsx
// DialogTrigger.tsx
import { Trigger as RadixTrigger } from '@radix-ui/react-dialog';

export const DialogTrigger = RadixTrigger;
DialogTrigger.displayName = 'DialogTrigger';
```

components/Dialog/Dialog.test.tsx
nickytonline: they do the same thing but this is a good opportunity to add an ESLint rule to prevent this. Want to tackle that @kepsteen or @Danielle254?

PR #368, opened 2025-04-25
test #181: add unit tests for Card component

### LINKED ISSUE #181: Create unit test for card.tsx

lillian-nguyen:
Introduces unit tests to verify expected behavior of Card component and its subcomponents. These tests cover: 

- rendering of component
- proper rendering of children
- prop forwarding behavior (tested with className)

closes #181


components/Card/cardContent.test.tsx
```diff
@@ -0,0 +1,59 @@
...
+import { CardContent } from './card'; 
+import React from 'react';
+
+describe('CardContent', () => {
+    it('renders CardContent with default styling', () => {
+        render(<CardContent data-testid="card-content"></CardContent>);
+
+        const cardContent = screen.getByTestId('card-content');
+
+        // Did CardContent render?
```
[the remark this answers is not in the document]
lillian-nguyen: Got it! Comments have been removed. 

Shashi wanted us to work towards Atomic design by decoupling function methods into separate files and creating unit tests for each.

components/Card/card.tsx
```diff
@@ -1,10 +1,15 @@
 // Copyright (c) Gridiron Survivor.
```
shashilo, line 1: This file need to be renamed with a uppercase C

components/Card/cardContent.test.tsx
```diff
@@ -0,0 +1,37 @@
+// Copyright (c) Gridiron Survivor.
```
shashilo, line 1: These test files should match their component file names. `CardContent.test.tsx`.

components/Card/cardHeader.test.tsx
```diff
@@ -0,0 +1,42 @@
...
+import React from 'react';
+
+describe('renders heading with default styling', () => {
+    it('renders CardHeader with default styling', () => {
+        render(<CardHeader data-testid="card-header">Card Header Content</CardHeader>)
+
+        const cardHeader = screen.getByTestId('card-header')
+
+        expect(cardHeader).toBeInTheDocument();
+        expect(cardHeader).toHaveClass('flex', 'flex-col', 'space-y-1.5 p-6');
```
shashilo, line 15: No need to test for these default styles that won't ever change. When we use Storybook, we'll add visual test that will cover for styling changes.

components/Card/cardHeader.test.tsx
```diff
@@ -0,0 +1,42 @@
+// Copyright (c) Gridiron Survivor.
+// Licensed under the MIT License.
+
+import { render, screen } from '@testing-library/react';
+import { CardHeader} from './CardHeader'; 
+import React from 'react';
+
+describe('renders heading with default styling', () => {
+    it('renders CardHeader with default styling', () => {
+        render(<CardHeader data-testid="card-header">Card Header Content</CardHeader>)
```
shashilo, line 10: Instead of adding the `data-testid` into the test file, add into the component directly. Then, it will always have the same id.

components/Card/cardHeader.test.tsx
```diff
@@ -0,0 +1,42 @@
+// Copyright (c) Gridiron Survivor.
+// Licensed under the MIT License.
+
+import { render, screen } from '@testing-library/react';
+import { CardHeader} from './CardHeader'; 
+import React from 'react';
+
+describe('renders heading with default styling', () => {
```
shashilo, line 8: The construction of this test needs work. The description should be `CardHeader Component`. Then, the `it`'s will describe the tests. Because this an Atom component, it should only have 1 describe in the file.

components/Card/cardHeader.test.tsx
```diff
@@ -0,0 +1,42 @@
...
+    it('renders CardHeader with default styling', () => {
```
shashilo, line 9: Once you set up the parent describe correctly, it will look like this as a test `CardHeader Component > renders with default styling`. Therefore, remove `CardHeader` from this description because the parent will already append it when testing.

components/Card/cardHeader.test.tsx
```diff
@@ -0,0 +1,42 @@
...
+        const ref = React.createRef<HTMLDivElement>();
+
+        render(<CardHeader ref={ref}>Ref test</CardHeader>);
+
+        expect(ref.current).toBeInstanceOf(HTMLDivElement);
+        expect(ref.current).toBeInTheDocument();
+    })
+})
+
+export {
```
shashilo, line 40: What's up with the extra lines? Is this a prettier format?

components/Card/cardHeader.test.tsx
```diff
@@ -0,0 +1,42 @@
...
+        render(<CardHeader data-testid="card-header" className="custom-class">Header Content</CardHeader>);
+
+        const cardHeader = screen.getByTestId('card-header');
+        expect(cardHeader).toHaveClass('custom-class')
+        expect(cardHeader).toHaveClass('flex', 'flex-col', 'space-y-1.5 p-6')
+    })
+})
+
+describe('CardHeader - forward Ref', () => {
+    it('forwards ref to DOM element (div)', () => {
```
shashilo, line 30: No need to test this

shashilo: We do need a test for custom attributes. The component test in `...props`. Therefore, `alt`, `aria-label`, etc could be passed in. We need to check that the component renders any custom attribute we pass as a prop.

shashilo (changes requested):

suzuhe481:
Just noticed I accidentally clicked Update Branch on this PR thinking it was mine. It shouldn't have affected anything, but just commenting to let everyone know.

kepsteen (dismissed):
LGTM, nice decoupling!

shashilo (approved):

PR #371, opened 2025-04-26
Hector/208-unit-test-for-radio-group

### LINKED ISSUE #208: Create unit test for radio-group.tsx

suzuhe481:
Closes #208

Creating initial unit tests for radio-group.
Tests the following components.
 - Radio Group 
- Radio Group Items

Run tests with 
```
pnpm run test ./components/RadioGroup/radio-group.test.tsx 
```

## Tests
- Tests the default state of each component, with any mandatory attributes (eg. value).
- Tests passing additional classNames.
- Tests rendering the component with no children.


kepsteen (changes requested):
Can you also test that the `<RadioGroupItem />` component functions as it's supposed to? The tests you have currently only test that it renders correctly under different conditions, but doesn't test the functionality of the radio item. What is this component used for?

suzuhe481:
> Can you also test that the `<RadioGroupItem />` component functions as it's supposed to? The tests you have currently only test that it renders correctly under different conditions, but doesn't test the functionality of the radio item. What is this component used for?

So for that, I should test that radio items get selected when clicked on?
Right now I can think of tests like...
 - Test that first item gets selected.
 - Test that last item gets selected.
 - When multiple items clicked, tests that the last item clicked gets selected.
 
Would these be enough?

I also assume I should make a radio group items with custom data-testid since I can't use the one in the app because it gets created by mapping the images and I can't add data-testids there.

kepsteen:
> > Can you also test that the `<RadioGroupItem />` component functions as it's supposed to? The tests you have currently only test that it renders correctly under different conditions, but doesn't test the functionality of the radio item. What is this component used for?
> 
> So for that, I should test that radio items get selected when clicked on? Right now I can think of tests like...
> 
> * Test that first item gets selected.
> * Test that last item gets selected.
> * When multiple items clicked, tests that the last item clicked gets selected.
> 
> Would these be enough?
> 
> I also assume I should make a radio group items with custom data-testid since I can't use the one in the app because it gets created by mapping the images and I can't add data-testids there.

<details>
<summary>Details</summary>



</details>

kepsteen:
Yes, test that when the item is clicked, it gets selected. If you select another item should the first item still be selected? Also, you should test that this component is accessible via the keyboard. You can use the `"@testing-library/user-event"` companion lib for this.

suzuhe481:
Added new tests to test for interactivity with mouse and keyboard.

package.json
Danielle254, line 1: why was this file changed?

components/RadioGroup/radio-group.test.tsx
```diff
@@ -0,0 +1,234 @@
...
+    expect(testChildElement).toBeInTheDocument();
+  });
+
+  it('renders radio group with additional custom classes', () => {
+    render(<RadioGroup className="custom-class-1 custom-class-2" />);
+
+    const radioGroupElement = screen.getByTestId('radio-group');
+
+    expect(radioGroupElement).toBeInTheDocument();
+    expect(radioGroupElement).toHaveClass('grid gap-2');
```
Danielle254, line 31: You've already tested this in the previous test. We can probably remove this line and just focus on the custom classes for this test.

components/RadioGroup/radio-group.test.tsx
```diff
@@ -0,0 +1,234 @@
...
+    expect(radioGroupElement).toHaveClass('custom-class-1 custom-class-2');
+  });
+
+  it('renders an empty radio group', () => {
```
Danielle254, line 35: is this desired behavior?

components/RadioGroup/radio-group.test.tsx
```diff
@@ -0,0 +1,234 @@
...
+
+    expect(radioGroupElement).toBeEmptyDOMElement();
+  });
+});
+
+describe('Radio Group Item Component', () => {
+  it('renders radio group item correctly with default classes and passed children and value', () => {
+    render(
+      <RadioGroupPrimitiveRoot>
+        <RadioGroupItem value="test value" aria-label="test" />
```
Danielle254, line 48: I'd need to review the shadcn documentation, but for a regular HTML radio group select element, you do not need to add `aria-label`s. The radio group item's value is read by the screen reader. We would only need an `aria-label` if the radio group item was say an icon or blank and not otherwise described by the value.

components/RadioGroup/radio-group.test.tsx
```diff
@@ -0,0 +1,234 @@
...
+    const radioGroupItemElement = screen.getByRole('radio', {
+      name: 'test',
+    });
+
+    expect(radioGroupItemElement).toBeEmptyDOMElement();
+  });
+});
+
+describe('Radio Group Item functionality', () => {
+  it('tests for first radio group items getting clicked', () => {
```
Danielle254, line 101: it('recognizes radio group items being clicked`)

components/RadioGroup/radio-group.test.tsx
```diff
@@ -0,0 +1,234 @@
...
+
+    const radioGroupItemElement = screen.getByRole('radio', {
+      name: 'test',
+    });
+
+    expect(radioGroupItemElement).toBeEmptyDOMElement();
+  });
+});
+
+describe('Radio Group Item functionality', () => {
```
Danielle254, line 100: does this need to be a separate test suite? I'd expect the respective tests to be in the appropriate component test suite.

components/RadioGroup/radio-group.test.tsx
suzuhe481: Removed the expect() statement that was used in a previous test.

components/RadioGroup/radio-group.test.tsx
suzuhe481: Removed this test since a RadioGroup should be expected to hold at least 1 RadioGroupItem, so an empty RadioGroup would be unnecessary to test.

components/RadioGroup/radio-group.test.tsx
suzuhe481: I used aria-labels as a replacement for data-testid, but didn't realize they were unnecessary.

I removed the aria-labels and how they were used to retrieve RadioGroupItems.

components/RadioGroup/radio-group.test.tsx
suzuhe481: Renamed some tests to follow this.

shashilo (changes requested):

package.json
```diff
@@ -49,16 +49,17 @@
...
     "@babel/preset-typescript": "^7.27.0",
-    "@storybook/test": "^8.6.12",
-    "@storybook/react": "^8.6.12",
-    "@storybook/nextjs": "^8.6.12",
-    "@storybook/blocks": "^8.6.12",
-    "@storybook/addon-onboarding": "^8.6.12",
-    "@storybook/addon-essentials": "^8.6.12",
     "@chromatic-com/storybook": "3",
     "@playwright/test": "^1.51.1",
+    "@storybook/addon-essentials": "^8.6.12",
```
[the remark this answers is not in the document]
suzuhe481: It was reordered when I installed @testing-library/user-event, but I didn't manually change it.

components/RadioGroup/radio-group.tsx
```diff
@@ -1,28 +1,29 @@
...
+import { forwardRef, ElementRef, ComponentPropsWithoutRef } from 'react';
 import * as RadioGroupPrimitive from '@radix-ui/react-radio-group';
 import { Circle } from 'lucide-react';
 
 import { cn } from '@/lib/utils';
 
-const RadioGroup = React.forwardRef<
-  React.ElementRef<typeof RadioGroupPrimitive.Root>,
-  React.ComponentPropsWithoutRef<typeof RadioGroupPrimitive.Root>
+const RadioGroup = forwardRef<
```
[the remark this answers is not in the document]
suzuhe481: I decoupled the components and updated their respective imports where they are used.

Danielle254 (dismissed):

package.json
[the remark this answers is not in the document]
nickytonline: I'm guessing they may were been manually added at one point, like copy pasted instead of added via `pnpm add` and adding the new package probably put them where they would be when running via `pnpm add`.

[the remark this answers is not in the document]
suzuhe481: I have a guess as to why it was reordered. I'm new to using pnpm, so I accidentally used ```pnpm install <package>``` instead of ```pnpm add <package>```. Your comment reminded me it uses ```add``` instead. And from what I'm reading online, even though ```pnpm install <package>``` works, it also scans and install packages from the lock file. So I think that when I ran it, it probably reinstalled the same packages while reordering them, while installing the one I specified.

hdkassamali (dismissed):
LGTM

components/RadioGroup/RadioGroup/RadioGroup.test.tsx
```diff
@@ -0,0 +1,31 @@
+// Copyright (c) Gridiron Survivor.
+// Licensed under the MIT License.
+
+import { RadioGroup } from './RadioGroup';
+import { render, screen } from '@testing-library/react';
+
+describe('Radio Group Component', () => {
```
shashilo, line 7: These test need to follow the capabilities of the Radio Group component. Here's the ShadCN docs as reference. Only look at the first example. The second one uses zod and the things. We're not conducting integration testing for this ticket. 

https://ui.shadcn.com/docs/components/radio-group

components/RadioGroup/RadioGroupItem/RadioGroupItem.test.tsx
```diff
@@ -0,0 +1,173 @@
...
+        <RadioGroupItem value="test value" />
+      </RadioGroupPrimitiveRoot>,
+    );
+
+    const radioGroupItemElement = screen.getByRole('radio');
+
+    expect(radioGroupItemElement).toBeEmptyDOMElement();
+  });
+
+  it('recognizes first radio group item being clicked', () => {
```
shashilo, line 57: This is an intergration test. Remove these. The task is just to add unit tests. Plus, we'd add the RadioGroup wrapper while using multiple RadioItems.

shashilo (changes requested):
The directories should not grouped. /ratioGroup/RatioGroup.tsx

components/RadioGroup/RadioGroupItem/RadioGroupItem.test.tsx
suzuhe481: I removed the integration tests and used RadioGroup as the wrapper instead of the previous RadioGroupPrimitiveRoot.

components/RadioGroup/RadioGroupItem/RadioGroupItem.test.tsx
```diff
@@ -0,0 +1,173 @@
...
+          value="test value"
+          className="custom-class-1 custom-class-2"
+        />
+      </RadioGroupPrimitiveRoot>,
+    );
+
+    const radioGroupItemElement = screen.getByRole('radio');
+
+    expect(radioGroupItemElement).toBeInTheDocument();
+    expect(radioGroupItemElement).toHaveClass(
```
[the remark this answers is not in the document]
suzuhe481: Removed the expect() lines checking for default classes.

components/RadioGroup/RadioGroup/RadioGroup.test.tsx
suzuhe481: Added a test to check for ShadCN's requirement of only button being checked at a time.

components/RadioGroup/RadioGroup.test.tsx
```diff
@@ -0,0 +1,60 @@
...
+        <RadioGroupItem value="option-2" data-testid="option-2" />
+      </RadioGroup>,
+    );
+
+    const radioGroupElement = screen.getByTestId('radio-group');
+    const optionOneElement = screen.getByTestId('option-1');
+    const optionTwoElement = screen.getByTestId('option-2');
+
+    expect(radioGroupElement).toBeInTheDocument();
+    expect(radioGroupElement).toHaveClass('grid gap-2');
```
nickytonline, line 22: refactor (nit): no need to check for layout styles.

components/RadioGroup/RadioGroup.test.tsx
```diff
@@ -0,0 +1,60 @@
...
+    expect(optionOneElement).toBeInTheDocument();
+    expect(optionTwoElement).toBeInTheDocument();
+  });
+
+  it('renders radio group with radio group items and no more than one item can be checked at a time', () => {
```
nickytonline, line 27: refactor (nit): Test names are a bit verbose. Shorter, behavior-focused titles make failure output easier to scan.  

e.g.

`renders radio group with radio group items and no more than one item can be checked at a time`  

would be better as

`only one item is checked at a time`

nickytonline (changes requested):
This is looking good. Some nits/change requests.

Also, assert the initial checked state before user interaction. Helps future-proof tests if defaults change.

components/RadioGroup/RadioGroup.test.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+    const radioGroupElement = screen.getByTestId('radio-group');
+    const optionOneElement = screen.getByTestId('option-1');
+    const optionTwoElement = screen.getByTestId('option-2');
+
+    expect(radioGroupElement).toBeInTheDocument();
+    expect(optionOneElement).toBeInTheDocument();
+    expect(optionTwoElement).toBeInTheDocument();
+
+    fireEvent.click(optionOneElement);
```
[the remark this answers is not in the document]
suzuhe481: Imported userEvent to replace fireEvent.

components/RadioGroup/RadioGroup.test.tsx
suzuhe481: Removed these style checks.

components/RadioGroup/RadioGroup.test.tsx
suzuhe481: Rewrote test name to be shorter.

alexappleget (approved):
LGTM

PR #380, opened 2025-04-30
Julianna/185-FeedbackView-Unit-Tests

### LINKED ISSUE #185: Create unit test for FeedbackView.tsx

mathematiCode:
Closes #185 
I added 9 unit tests and some comments because I was a little unsure about how to test certain aspects of this. 
image
Like my other unit test PR, I wanted to add a test that tests if the buttons' hover state/background color is different than the default, but I wasn't able to get that to pass. https://discord.com/channels/964903664265359430/1366570754933657620

### DIFF
--- a/components/FeedbackView/FeedbackView.test.tsx
+++ b/components/FeedbackView/FeedbackView.test.tsx
@@ -45,7 +45,6 @@ describe('FeedbackView', () => {
 
   const renderFeedbackView = () => {
     return render(
-      <div data-testid="feedback-view">
         <FeedbackView
           allGiftSuggestions={[]}
           budget={''}
@@ -54,7 +53,6 @@ describe('FeedbackView', () => {
           onGiftUpdate={mockOnGiftUpdate}
           recipient={mockProfile}
         />
-      </div>
     );
   };
 
--- a/components/FeedbackView/FeedbackView.test.tsx
+++ b/components/FeedbackView/FeedbackView.test.tsx
@@ -1,26 +1,139 @@
 import { render, screen } from '@testing-library/react';
-import  FeedbackView  from './FeedbackView';
+import { GiftSuggestion } from '@/app/types/giftSuggestion';
+import { Profile } from '@/app/types/profile';
+import FeedbackView from './FeedbackView';
 
 describe('FeedbackView', () => {
+  const mockGiftSuggestion: GiftSuggestion = {
+    id: '1',
+    title: 'Test Gift',
+    price: '100',
+    description: 'Test Description',
+    matchReasons: ['test reason'],
+    matchScore: 0.8,
+    imageUrl: 'test.jpg',
+  };
+
+  const mockProfile: Profile = {
+    id: '1',
+    display_name: 'Test User',
+    email: 'test@example.com',
+    categories: ['test'],
+    created_at: new Date().toISOString(),
+    updated_at: new Date().toISOString(),
+    onboarding_complete: true,
+  };
+
   it('should render', () => {
-    render(<FeedbackView />);
+    render(
+      <FeedbackView
+        allGiftSuggestions={[]}
+        budget={''}
+        gift={mockGiftSuggestion}
+        handleFeedback={() => {}}
+        onGiftUpdate={() => {}}
+        recipient={null}
+      />
+    );
   });
+
+  describe('Loading State', () => {
+    it('should show loading spinner and hide buttons when isLoading is true', () => {
+      const { rerender } = render(
+        <FeedbackView
+          allGiftSuggestions={[]}
+          budget={''}
+          gift={mockGiftSuggestion}
+          handleFeedback={() => {}}
+          onGiftUpdate={() => {}}
+          recipient={mockProfile}
+        />
+      );
+
+      // Initially, loading spinner should not be visible
+      expect(screen.queryByRole('status')).not.toBeInTheDocument();
+      
+      // Buttons should be visible
+      expect(screen.getByText('Too Expensive')).toBeInTheDocument();
+      expect(screen.getByText('Not Their Style')).toBeInTheDocument();
+      expect(screen.getByText('They Might Have This')).toBeInTheDocument();
+
+      // Simulate loading state by clicking a button
+      const button = screen.getByText('Too Expensive').closest('button');
+      if (button) {
+        button.click();
+      }
+
+      // After clicking, loading spinner should be visible
+      expect(screen.getByRole('status')).toBeInTheDocument();
+      
+      // Buttons should not be visible
+      expect(screen.queryByText('Too Expensive')).not.toBeInTheDocument();
+      expect(screen.queryByText('Not Their Style')).not.toBeInTheDocument();
+      expect(screen.queryByText('They Might Have This')).not.toBeInTheDocument();
+    });
+
+    it('should show buttons and hide loading spinner when isLoading is false', () => {
+      render(
+        <FeedbackView
+          allGiftSuggestions={[]}
+          budget={''}
+          gift={mockGiftSuggestion}
+          handleFeedback={() => {}}
+          onGiftUpdate={() => {}}
+          recipient={mockProfile}
+        />
+      );
+
+      // Initially, loading spinner should not be visible
+      expect(screen.queryByRole('status')).not.toBeInTheDocument();
+      
+      // Buttons should be visible
+      expect(screen.getByText('Too Expensive')).toBeInTheDocument();
+      expect(screen.getByText('Not Their Style')).toBeInTheDocument();
+      expect(screen.getByText('They Might Have This')).toBeInTheDocument();
+    });
+  });
+
+  // if isLoading is true, no buttons should be rendered and loading spinner should be rendered
+  // if isLoading is false, 3 buttons should be rendered and "Give Us Feedback" and the Chevron should be rendered 
+  // when a button is clicked, the handleFeedbackSubmit function should be called
+  // when a button is clicked, the handleFeedbackSubmit function should be called with the correct argument
+  // when a button is clicked, isLoading should switch to true (it may take a second though) and then back to false after it loads.
+  // each button should have a title and a subtitle 
+  // each title and corresponding subtitle should be from the same index in the buttonVariants array
+  // there should exist a button with each title and subtitle. 
+  // the loading spinner and buttons should never be rendered at the same time
+  // the items in a flexCol container should have matching starting x values (be vertically aligned)
+  // the 3 buttons should have matching starting x values (be vertically aligned)
+  // if the user clicks one of the buttons, onGiftUpdate should be called
+  // if the user clicks one of the buttons, updatedGift should be initialized and not equal to gift.
+
+  // Organized Test Groups:
+
+  // Loading State Tests:
+  // - Test loading state visibility (spinner vs buttons)
+  // - Test loading state transitions
+  // - Test mutual exclusivity of spinner and buttons
+
+  // Button Click Behavior Tests:
+  // - Test handleFeedbackSubmit function calls
+  // - Test correct argument passing
+  // - Test loading state changes during button clicks
+
+  // Button Content Tests:
+  // - Test presence of title and subtitle
+  // - Test correct buttonVariants array mapping
+  // - Test completeness of button set
+
+  // Layout Tests:
+  // - Test vertical alignment of flexCol items
+  // - Test vertical alignment of buttons
+
+  // Gift Update Tests:
+  // - Test onGiftUpdate function calls
+  // - Test updatedGift initialization and changes
 });
-
-
-// if isLoading is true, no buttons should be rendered and loading spinner should be rendered
-// if isLoading is false, 3 buttons should be rendered and "Give Us Feedback" and the Chevron should be rendered 
-// when a button is clicked, the handleFeedbackSubmit function should be called
-// when a button is clicked, the handleFeedbackSubmit function should be called with the correct argument
-// when a button is clicked, isLoading should switch to true (it may take a second though) and then back to false after it loads.
-// each button should have a title and a subtitle 
-// each title and corresponding subtitle should be from the same index in the buttonVariants array
-// there should exist a button with each title and subtitle. 
-// the loading spinner and buttons should never be rendered at the same time
-// the items in a flexCol container should have matching starting x values (be vertically aligned)
-// the 3 buttons should have matching starting x values (be vertically aligned)
-// if the user clicks one of the buttons, onGiftUpdate should be called
-// if the user clicks one of the buttons, updatedGift should be initialized and not equal to gift.
 
 
 
--- a/components/FeedbackView/FeedbackView.test.tsx
+++ b/components/FeedbackView/FeedbackView.test.tsx
@@ -24,22 +24,9 @@ describe('FeedbackView', () => {
     onboarding_complete: true,
   };
 
-  it('should render', () => {
-    render(
-      <FeedbackView
-        allGiftSuggestions={[]}
-        budget={''}
-        gift={mockGiftSuggestion}
-        handleFeedback={() => {}}
-        onGiftUpdate={() => {}}
-        recipient={null}
-      />
-    );
-  });
-
-  describe('Loading State', () => {
-    it('should show loading spinner and hide buttons when isLoading is true', () => {
-      const { rerender } = render(
+  const renderFeedbackView = () => {
+    return render(
+      <div data-testid="feedback-view">
         <FeedbackView
           allGiftSuggestions={[]}
           budget={''}
@@ -48,24 +35,40 @@ describe('FeedbackView', () => {
           onGiftUpdate={() => {}}
           recipient={mockProfile}
         />
-      );
+      </div>
+    );
+  };
+
+  it('should render', () => {
+    renderFeedbackView();
+    expect(screen.getByTestId('feedback-view')).toBeInTheDocument();
+  });
+
+  
+
+  describe('Loading State', () => {
+    it('should show loading spinner and hide buttons when isLoading is true', () => {
+      const { rerender } = renderFeedbackView();
 
       // Initially, loading spinner should not be visible
-      expect(screen.queryByRole('status')).not.toBeInTheDocument();
+      expect(screen.queryByTestId('loading-spinner')).not.toBeInTheDocument();
       
       // Buttons should be visible
-      expect(screen.getByText('Too Expensive')).toBeInTheDocument();
-      expect(screen.getByText('Not Their Style')).toBeInTheDocument();
-      expect(screen.getByText('They Might Have This')).toBeInTheDocument();
+      const expensiveButton = screen.getByText('Too Expensive').closest('button');
+      const styleButton = screen.getByText('Not Their Style').closest('button');
+      const haveButton = screen.getByText('They Might Have This').closest('button');
+
+      expect(expensiveButton).toBeInTheDocument();
+      expect(styleButton).toBeInTheDocument();
+      expect(haveButton).toBeInTheDocument();
 
       // Simulate loading state by clicking a button
-      const button = screen.getByText('Too Expensive').closest('button');
-      if (button) {
-        button.click();
+      if (expensiveButton) {
+        expensiveButton.click();
       }
 
       // After clicking, loading spinner should be visible
-      expect(screen.getByRole('status')).toBeInTheDocument();
+      expect(screen.getByTestId('loading-spinner')).toBeInTheDocument();
       
       // Buttons should not be visible
       expect(screen.queryByText('Too Expensive')).not.toBeInTheDocument();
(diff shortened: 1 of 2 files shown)


components/FeedbackView/FeedbackView.test.tsx
```diff
@@ -0,0 +1,260 @@
+import { render, screen, waitFor, act } from '@testing-library/react';
+import { GiftSuggestion } from '@/app/types/giftSuggestion';
+import { Profile } from '@/app/types/profile';
+import FeedbackView from './FeedbackView';
+import userEvent from '@testing-library/user-event';
+import { generateAndUpdateNewGiftSuggestion } from '@/lib/generateAndUpdateNewGiftSuggestion';
+
+// I'm wrapping all of the tests that click one of the feedback buttons in act() because they all set state and I was getting a lot of errors that anything that triggered a state change needs to be wrapped in act().
```
kepsteen, line 8: I'm getting this error when I run the tests: 
`console.error
    Warning: The current testing environment is not configured to support act(...)`
    
   Are you seeing this as well?

kepsteen (changes requested):

components/FeedbackView/FeedbackView.test.tsx
mathematiCode: Yes but I thought I fixed it. I will investigate further.

mathematiCode: @kepsteen I fixed it now!

components/FeedbackView/FeedbackView.test.tsx
```diff
@@ -0,0 +1,260 @@
...
+    });
+
+    await waitFor(() => {
+      expect(screen.queryByTestId('loading-spinner')).not.toBeInTheDocument();
+      expect(screen.getByTestId('feedback-button-0')).toBeInTheDocument();
+    });
+  });
+
+
+  it('Each button renders with the correct title and corresponding subtitle', () => {
```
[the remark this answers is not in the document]
mathematiCode: @kepsteen Done!

components/FeedbackView/FeedbackView.test.tsx
mathematiCode, line 1: I'm wrapping all of the tests that click one of the feedback buttons in act() because they all set state and I was getting a lot of errors that anything that triggered a state change needs to be wrapped in act().

components/FeedbackView/FeedbackView.test.tsx
```diff
@@ -0,0 +1,254 @@
...
+      { title: 'They Might Have This', subtitle: 'Show alternatives' },
+    ];
+
+    buttonVariants.forEach(({ title, subtitle }) => {
+      const button = screen.getByText(title).closest('button');
+      expect(button).toBeInTheDocument();
+      expect(button).toHaveTextContent(subtitle);
+    });
+  });
+
```
mathematiCode, line 100: I can't access handleFeedbackSubmit directly since it's not a prop so this tests whether onGiftUpdate is called as a result of handleFeedbackSubmit running.

components/FeedbackView/FeedbackView.test.tsx
```diff
@@ -0,0 +1,254 @@
...
+    await waitFor(() => {
+      expect(consoleErrorSpy).toHaveBeenCalledWith('Failed to update gift suggestion');
+    });
+
+    consoleErrorSpy.mockRestore();
+
+    expect(mockOnGiftUpdate).not.toHaveBeenCalled();
+  });
+
+  // I'm not entirely sure if this is even testing whether an error is thrown or not because it passes even when I just use my renderFeedbackView() function without using the mockOnGiftUpdate function with the mockRejectedValue. I'm not able to get it to pass any other way and still test that it throws an error. When I check test coverage it says the only line not covered is line 42 (which throws this error)
```
mathematiCode, line 227: I'm not entirely sure if this is even testing whether an error is thrown or not because it passes even when I just use my renderFeedbackView() function without using the mockOnGiftUpdate function with the mockRejectedValue. I'm not able to get it to pass any other way and still test that it throws an error. When I check test coverage it says the only line not covered is line 42 (which throws this error)

components/FeedbackView/FeedbackView.test.tsx
```diff
@@ -0,0 +1,245 @@
...
+    await waitFor(() => {
+      expect(consoleErrorSpy).toHaveBeenCalledWith('Failed to update gift suggestion');
+    });
+
+    consoleErrorSpy.mockRestore();
+
+    expect(mockOnGiftUpdate).not.toHaveBeenCalled();
+  });
+
+  it('When onGiftUpdate fails, an error is thrown', async () => {
```
mathematiCode, line 219: I'm not entirely sure if this is even testing whether an error is thrown or not because it passes even when I just use my renderFeedbackView() function without using the mockOnGiftUpdate function with the mockRejectedValue. I'm not able to get it to pass any other way and still test that it throws an error. When I check test coverage it says the only line not covered is line 42 (which throws this error)

kepsteen: I'm not sure either maybe @shashilo or @nickytonline could weigh in? I think everything else looks good, so I can approve after they give their feedback.

mathematiCode: image
@kepsteen I made some adjustments and checked that it is actually testing that an error is thrown by creating another test that should fail on purpose (and then I deleted it). So now I have more confidence that this is testing the right thing.

components/FeedbackView/FeedbackView.tsx
```diff
@@ -64,25 +64,27 @@ const FeedbackView = ({
 
 
   return (
-    <div className="flex flex-col m-4 h-full">
+    <div className="flex flex-col m-4 h-full" data-testid="feedback-view">
       {isLoading ? (
         <>
```
shashilo, line 69: Can these be wrapped up in elements instead of fragments? Then, you could move the data-testid to the parent wrapper.

components/FeedbackView/FeedbackView.tsx
```diff
@@ -64,25 +64,27 @@ const FeedbackView = ({
...
           <div className="flex justify-center items-center h-full">
-            <div className="border-t-4 border-red-500 border-solid w-12 h-12 rounded-full animate-spin" />
+            <div data-testid="loading-spinner" className="border-t-4 border-red-500 border-solid w-12 h-12 rounded-full animate-spin" />
           </div>
         </>
       ) : (
         <>
```
shashilo, line 75: Same here. We just need to test that the condition renders one or the other.

components/FeedbackView/FeedbackView.tsx
```diff
@@ -64,25 +64,27 @@ const FeedbackView = ({
...
           <ChevronLeft
+            data-testid="back-chevron"
             className="hover:cursor-pointer"
             onClick={handleFeedback}
           />
-          <h1 className="text-sm text-[#21443D] font-bold mx-auto mt-4">
+          <h1 data-testid="feedback-title" className="text-sm text-[#21443D] font-bold mx-auto mt-4">
```
shashilo, line 81: Remove this once you set the parent wrapper up with a data-testid

components/FeedbackView/FeedbackView.tsx
```diff
@@ -64,25 +64,27 @@ const FeedbackView = ({
...
             Give Us Feedback
           </h1>
           <div className="flex flex-col justify-center mt-4 gap-4">
             {buttonVariants.map(({ title, subtitle, id }) => (
```
shashilo, line 85: Let's decouple this logic into a new component. <FeedbackViewButtons>. This is a great case to move the logic out. This also will allow for isolated tests.

components/FeedbackView/FeedbackView.test.tsx
```diff
@@ -0,0 +1,250 @@
+import { render, screen, waitFor, act } from '@testing-library/react';
+import { GiftSuggestion } from '@/app/types/giftSuggestion';
+import { Profile } from '@/app/types/profile';
+import FeedbackView from './FeedbackView';
+import userEvent from '@testing-library/user-event';
+import { generateAndUpdateNewGiftSuggestion } from '@/lib/generateAndUpdateNewGiftSuggestion';
+
+jest.mock('@/lib/generateAndUpdateNewGiftSuggestion', () => ({
+  generateAndUpdateNewGiftSuggestion: jest.fn().mockResolvedValue({
```
shashilo, line 9: The return of this needs to be adjusted. Because the property that is passed into this component is passed into this function, it expects that what you pass into it, should return something matching it. This is returning just a random object that doesn't match the properties passed into the initial component. This might be difficult to read, so let me know.

components/FeedbackView/FeedbackView.tsx
```diff
@@ -64,25 +64,27 @@ const FeedbackView = ({
...
               <button
+                data-testid={`feedback-button-${id}`}
                 type="button"
                 key={id}
```
shashilo, line 89: Change this to a computer name. Then, you can use it for both the data-testid and the ID Having it be simple numbers may cause identifying issues down the road.

components/FeedbackView/FeedbackView.test.tsx
```diff
@@ -0,0 +1,250 @@
...
+    expect(haveButton).toBeInTheDocument();
+
+    expect(screen.getByTestId('feedback-title')).toBeInTheDocument();
+    expect(screen.getByTestId('back-chevron')).toBeInTheDocument();
+  });
+
+  it('Each button renders with the correct title and corresponding subtitle', () => {
+    renderFeedbackView();
+
+    const buttonVariants = [
```
shashilo, line 86: Why are we resetting the button object here instead of using the variable set in the component already?

components/FeedbackView/FeedbackView.test.tsx
```diff
@@ -0,0 +1,250 @@
...
+  beforeEach(() => {
+    jest.clearAllMocks();
+  });
+
+  it('renders the feedback interface', () => {
+    renderFeedbackView();
+    expect(screen.getByTestId('feedback-view')).toBeInTheDocument();
+  });
+
+  it('When isLoading is false, the loading spinner is hidden and the buttons are visible', () => {
```
shashilo, line 66: Add the parent elements with BEM class naming convention. Then, you won't need to check for all these elements. You just need to look at the parent if they exist or not.

components/FeedbackView/FeedbackView.test.tsx
```diff
@@ -0,0 +1,250 @@
...
+          description: 'New Test Description',
+          matchReasons: ['new test reason'],
+          matchScore: 0.9,
+          imageUrl: 'new-test.jpg',
+        }),
+      );
+    });
+  });
+
+  it('All buttons are focusable, tabbable, and clickable.', async () => {
```
shashilo, line 167: Don't need this test. The buttons should be created from our UI component library. Then, when we add A11y test to those components, these individual tests won't be needed.

components/FeedbackView/FeedbackView.test.tsx
```diff
@@ -0,0 +1,250 @@
...
+        'Failed to update gift suggestion',
+      );
+    });
+
+    consoleErrorSpy.mockRestore();
+
+    expect(mockOnGiftUpdate).not.toHaveBeenCalled();
+  });
+
+  it('When onGiftUpdate fails, an error is thrown', async () => {
```
shashilo, line 226: If it's in the code, you test it. In this case, we're throwing an error when something happens. So we would write a test to cover the lines of code.

shashilo (changes requested):

components/FeedbackView/FeedbackView.tsx
mathematiCode: I wasn't sure exactly which part you wanted to be extracted out. So I made a ButtonGroup component that takes in the variants array.

components/FeedbackView/FeedbackView.tsx
mathematiCode: Done!

components/FeedbackView/FeedbackView.test.tsx
mathematiCode: Okay I deleted it. Thanks!

components/FeedbackView/FeedbackView.test.tsx
mathematiCode: Lots of discussion around this here: https://discord.com/channels/964903664265359430/1367204236798197841/1379881544218443806
I'm still not entirely sure how I should fix this.

components/FeedbackView/FeedbackView.test.tsx
mathematiCode: Okay thanks!

kepsteen (approved):

components/FeedbackView/FeedbackView.tsx
```diff
@@ -63,36 +77,31 @@ const FeedbackView = ({
...
+    <div className="flex flex-col m-4 h-full" data-testid="feedback-view">
       {isLoading ? (
-        <>
-          <div className="flex justify-center items-center h-full">
-            <div className="border-t-4 border-red-500 border-solid w-12 h-12 rounded-full animate-spin" />
-          </div>
-        </>
+        <div
+          className="flex justify-center items-center h-full"
+          data-testid="feedback-view--loading"
```
shashilo, line 84:
```suggestion
          data-testid="feedbackView__loading"
```
The `--` are for modifiers. This is a child element.

components/FeedbackView/FeedbackView.tsx
```diff
@@ -63,36 +77,31 @@ const FeedbackView = ({
...
+        >
+          <div className="border-t-4 border-red-500 border-solid w-12 h-12 rounded-full animate-spin" />
+        </div>
       ) : (
-        <>
+        <div
+          className="flex flex-col m-4 h-full"
+          data-testid="feedback-view--loaded"
```
shashilo, line 91:
```suggestion
          data-testid="feedbackView__loading"
```
The `--` are for modifiers. This is a child element.

components/FeedbackView/FeedbackView.test.tsx
shashilo: You should mock one of the returns from OpenAI. Like the exact response object. Then, this response will be accurate for a unit test.

components/FeedbackView/ButtonGroup.tsx
```diff
@@ -0,0 +1,43 @@
...
+const ButtonGroup = ({
+  variants,
+  handleFeedbackSubmit,
+}: {
+  variants: { title: string; subtitle: string; id: string }[];
+  handleFeedbackSubmit: (feedback: string) => void;
+}): JSX.Element => {
+  return (
+    <div
+      data-testid="button-group"
```
shashilo, line 22:
```suggestion
      data-testid="buttonGroup"
```

components/FeedbackView/ButtonGroup.tsx
```diff
@@ -0,0 +1,43 @@
+// Copyright (c) Gridiron Survivor.
```
shashilo, line 1: Is this button group reusable throughout the app or just for FeedbackView page?

components/FeedbackView/ButtonGroup.tsx
```diff
@@ -0,0 +1,43 @@
...
+ * @param {object[]} props.variants - The list of button variants to render
+ * @param {Function} props.handleFeedbackSubmit - The function to handle the feedback submit
+ * @returns {JSX.Element} The rendered button group component
+ */
+const ButtonGroup = ({
+  variants,
+  handleFeedbackSubmit,
+}: {
+  variants: { title: string; subtitle: string; id: string }[];
+  handleFeedbackSubmit: (feedback: string) => void;
```
shashilo, line 18: interface Variant {
  title: string;
  subtitle: string;
  id: string;
}

interface ButtonGroupProps {
  variants: Variant[];
  handleFeedbackSubmit: (feedback: string) => void;
}

const ButtonGroup = ({ variants, handleFeedbackSubmit }: ButtonGroupProps) => {
  // component logic here
};

components/FeedbackView/ButtonGroup.tsx
```diff
@@ -0,0 +1,43 @@
...
+}): JSX.Element => {
+  return (
+    <div
+      data-testid="button-group"
+      className="flex flex-col justify-center mt-4 gap-4"
```
shashilo, line 23:
```suggestion
      className="buttonGroup flex flex-col justify-center mt-4 gap-4"
```

components/FeedbackView/FeedbackView.test.tsx
shashilo: Console.error won't trigger a failure in the app. You need to `throw new Error()`

components/FeedbackView/ButtonGroup.test.tsx
```diff
@@ -0,0 +1,31 @@
+import { render, screen } from '@testing-library/react';
+import ButtonGroup from './ButtonGroup';
+
+it('Each button renders with the correct title and corresponding subtitle', () => {
```
shashilo, line 4: Need to test that the number of objects we pass in match the number displayed.

shashilo (changes requested):

components/FeedbackView/ButtonGroup.tsx
mathematiCode: @shashilo It's not really reusable as it's currently set up. The way that handleFeedbackSubmit is handled is pretty specific to the FeedbackView component. Other use cases for a group of buttons wouldn't need to pass in the variant title and subtitle interpolated like this. 
image
And they might not need to display the variant title and subtitle the same way.

components/FeedbackView/FeedbackView.test.tsx
mathematiCode: @shashilo Do we want this scenario to trigger a failure in the app? I got this test to pass asserting that an error is shown in the console.

components/FeedbackView/ButtonGroup.test.tsx
mathematiCode: @shashilo Do we want this scenario to trigger a failure in the app? I got this test to pass asserting that an error is shown in the console.

mathematiCode: @shashilo I added a test for how to make sure the number of items passed in matches the number displayed. I also tried changing the console.error to throw new Error, but now my test fails. So I just want to confirm, you want me to change both the actual component code to trigger a failure and the unit test to assert that failure?

components/FeedbackView/ButtonGroup.test.tsx
```diff
@@ -0,0 +1,56 @@
...
```
nickytonline, line 4: phrasing (nit): for tests it should read e.g. 

```suggestion
it('Renders each button with the correct title and corresponding subtitle', () => {
```

would read as `It renders each button with the correct title and corresponding subtitle`, instead of `It each button renders with the correct title and corresponding subtitle`

components/FeedbackView/ButtonGroup.test.tsx
```diff
@@ -0,0 +1,56 @@
...
+  );
+
+  buttonVariants.forEach(({ title, subtitle }) => {
+    const button = screen.getByText(title).closest('button');
+    expect(button).toBeInTheDocument();
+    expect(button).toHaveTextContent(subtitle);
+  });
+});
+
+it('The correct number of buttons are rendered based on how many are passed in', () => {
```
nickytonline, line 33: phrasing (nit): similar to above but also too much info in the described message. Consider:

```suggestion
it('Renders the correct number of buttons', () => {
```

components/FeedbackView/ButtonGroup.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+interface ButtonGroupProps {
+  variants: Variant[];
+  handleFeedbackSubmit: (feedback: string) => void;
+}
+
+/**
+ * Component that renders a group of buttons.
+ * @param {object} props - The component props
+ * @param {object[]} props.variants - The list of button variants to render
```
nickytonline, line 20: refactor (blocking): use proper type

```suggestion
 * @param {Variant[]} props.variants - The list of button variants to render
```

components/FeedbackView/ButtonGroup.tsx
```diff
@@ -0,0 +1,51 @@
...
+ * @param {Function} props.handleFeedbackSubmit - The function to handle the feedback submit
```
nickytonline, line 21: refactor (blocking): use the proper types

```suggestion
 * @param {Variant[]} props.variants - The list of button variants to render
 * @param {(feedback: string) => void;} props.handleFeedbackSubmit - The function to handle the feedback submit
```

You can also create a type for the feedback callback and use that in both places, e.g.

```typescript
type FeedbackCallback = (feedback: string) => void;
```

nickytonline (dismissed):

shashilo (approved):

PR #379, opened 2025-04-30
Julianna/201-LoginButton-Unit-Tests

mathematiCode:
Closes #201 
I added 5 unit tests for this component to test all (hopefully) intended functionality. 
I did my best to avoid checking implementation details like if isLoading is set to true or false. 
I tried to add a test to test if the background color on hover was different than the default color, but I wasn't able to get this passing due to the way tailwind classes are applied (I think?). 
image

Issue #154, opened 2025-03-13
Create unit test for SnowOverlayProvider.tsx


Danielle254:
perhaps integration test also

PR #381, opened 2025-04-30
Zoobia/212 unit test for snow overlay toggle

zoobiasyyed:
First attempt at a unit test! Kharram (i think thats how you spell his name) helped me out a bit, I had trouble figuring out why one of my tests werent passing. 

The point of this unit test is to confirm if the sun and snowflake icons were toggling with each click. The initial error I recieved I now realize was because in my unit test I was rendering the <SnowOverToggle/> component without wrapping the <SnowOverlayProvider/>. 

hopefully I did okay for the first attempt!

### DIFF
--- a/components/SnowOverlayToggle/SnowOverlayToggle.test.tsx
+++ b/components/SnowOverlayToggle/SnowOverlayToggle.test.tsx
@@ -1,5 +1,0 @@
-// Copyright (c) Gridiron Survivor.
-// Licensed under the MIT License.
-
-import { render, screen } from '@testing-library/react';
-import SnowOverlayToggle from './SnowOverlayToggle';
--- a/components/SnowOverlayToggle/SnowOverlayToggle.test.tsx
+++ b/components/SnowOverlayToggle/SnowOverlayToggle.test.tsx
@@ -1,15 +1,25 @@
 // Copyright (c) Gridiron Survivor.
 // Licensed under the MIT License.
 
+
 import { SnowOverlayToggle } from "./SnowOverlayToggle";
 import { fireEvent, render, screen } from '@testing-library/react';
+import { SnowOverlayProvider } from "@/providers/SnowOverlayProvider";
 
 
 describe('SnowOverlayToggle', () => {
   it('Toggles the snowfall overlay', ()=> {
-    render(<SnowOverlayToggle/>)
-    expect(screen.queryByTestId('snow-icon')).toBeInTheDocument();
+    render(
+      <SnowOverlayProvider>
+        <SnowOverlayToggle />
+      </SnowOverlayProvider>,
+    );
+    const sunIcon = screen.getByTestId('sun-icon');
+    expect(sunIcon).toBeInTheDocument();
+
     fireEvent.click(screen.getByTestId('snow-overlay-toggle'));
-    expect(screen.queryByTestId('sun-icon')).toBeInTheDocument();
+
+    const snowIcon = screen.getByTestId('snow-icon');
+    expect(snowIcon).toBeInTheDocument();
   })
 })
--- a/components/SnowOverlayToggle/SnowOverlayToggle.test.tsx
+++ b/components/SnowOverlayToggle/SnowOverlayToggle.test.tsx
@@ -1,0 +1,15 @@
+// Copyright (c) Gridiron Survivor.
+// Licensed under the MIT License.
+
+import { SnowOverlayToggle } from "./SnowOverlayToggle";
+import { fireEvent, render, screen } from '@testing-library/react';
+
+
+describe('SnowOverlayToggle', () => {
+  it('Toggles the snowfall overlay', ()=> {
+    render(<SnowOverlayToggle/>)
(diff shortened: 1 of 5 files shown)


components/SnowOverlayToggle/SnowOverlayToggle.test.tsx
```diff
@@ -0,0 +1,25 @@
...
+        <SnowOverlayToggle />
+      </SnowOverlayProvider>,
+    );
+    const sunIcon = screen.getByTestId('sun-icon');
+    expect(sunIcon).toBeInTheDocument();
+
+    fireEvent.click(screen.getByTestId('snow-overlay-toggle'));
+
+    const snowIcon = screen.getByTestId('snow-icon');
+    expect(snowIcon).toBeInTheDocument();
```
mathematiCode, line 23: It looks like you're testing whether the toggle actually toggles, but not whether toggling it triggers the SnowOverlay to appear and disappear. I guess the closest you can get to testing that without going outside of this file is to test whether the value of isSnowing is switching to the correct value or not.

components/SnowOverlayToggle/SnowOverlayToggle.test.tsx
```diff
@@ -0,0 +1,25 @@
...
+  it('Toggles the snowfall overlay', ()=> {
+    render(
+      <SnowOverlayProvider>
+        <SnowOverlayToggle />
+      </SnowOverlayProvider>,
+    );
+    const sunIcon = screen.getByTestId('sun-icon');
+    expect(sunIcon).toBeInTheDocument();
+
+    fireEvent.click(screen.getByTestId('snow-overlay-toggle'));
```
mathematiCode, line 20: From my research it seems generally preferred to use userEvent over fireEvent since [it simulates a more realistic user interaction](https://testing-library.com/docs/user-event/intro/). I'm not sure if we have a preference for this project. And since toggling in your case looks like it will cause state changes through useSnowOverlay() you will probably have to wrap it in act() like this:
```
 await act(async () => {
               await userEvent.click(putYourToggleHere);
  }); 
```

components/SnowOverlayToggle/SnowOverlayToggle.test.tsx
mathematiCode: @zoobiasyyed So I realized testing the value of isSnowing would be testing the implementation details, which is a bad practice, so never mind about that. I learned that it is possible to include the layout that's calling the SnowOverlay in the render function of your test and then test that the SnowOverlay actually appears and disappears, but I think that would make more sense as part of the [SnowOverlayProvider unit/integration tests](https://github.com/LetsGetTechnical/elecretanta/issues/154). So I think the way you have this is fine.

mathematiCode (changes requested):
We decided to use userEvent for all user-triggered actions because it more closely resembles an actual user engaging with the components. So can you change all of your fireEvents to userEvents? Everything else looks good! I used it a bunch in this PR if you need a reference: https://github.com/LetsGetTechnical/elecretanta/pull/379/files

components/SnowOverlayToggle/SnowOverlayToggle.test.tsx
zoobiasyyed: just changed my fireEvent to userEvent!

mathematiCode: @zoobiasyyed It looks like at least one of your tests is failing now. Probably due to it being asyncronous. Try putting `await` in front of the line userEvent.click(...

zoobiasyyed: @mathematiCode thank you! I think it should work now

components/SnowOverlayToggle/SnowOverlayToggle.test.tsx
```diff
@@ -0,0 +1,26 @@
...
+import { SnowOverlayProvider } from "@/providers/SnowOverlayProvider";
+import { userEvent } from "@storybook/test";
+
+
+describe('SnowOverlayToggle', () => {
+  it('Toggles the snowfall overlay', async ()=> {
+    render(
+      <SnowOverlayProvider>
+        <SnowOverlayToggle />
+      </SnowOverlayProvider>,
```
nickytonline, line 16: refactor (blocking): the test is for the SnowOverlayToggle component, not the SnowOverlayToggle and SnowOverlayProvider.

Mock the SnowOverlayProvider in the tests and omit it from the render since we're focusing on unit tests at the moment for the tests for components. Having the provider is more of an integration test.

I'll defer to @shashilo though as I'm not sure of all the testing guidelines he provided to the team.

```suggestion
      <SnowOverlayToggle />
```

nickytonline (changes requested):

bethanyann:
If this pull request is for a task you were assigned, can you add the link to that task in the pull request description?  

It's helpful when reviewing pull requests to know which ticket it's related to and have that link present in the description!

shashilo (dismissed):

nickytonline (approved):
:shipit:

PR #382, opened 2025-05-01
test: [#214] add unit test for textarea

### LINKED ISSUE #214: Create unit test for textarea.tsx

CooleyWC:
Add 3 tests:

1. Checks if the Textarea renders.
2. Checks if the Textarea renders with custom className.
3. Checks if the Textarea renders with props.

Closes #214


kepsteen (changes requested):
You've tested that this component is rendering as it should be, but you also need to test that it functionally does what it's supposed to. Think about how this component is being used in the application? 

We want to ensure through testing that we have the confidence that this component is behaving how it should be.

CooleyWC:
Added a test that checks that text can be entered into the Textarea component. Let me know if there's anything else I need to add or adjust.

components/TextArea/textarea.test.tsx
```diff
@@ -0,0 +1,36 @@
...
+            />)
+        const textarea = screen.getByRole("textbox")
+        expect(textarea).toHaveAttribute("placeholder", "test-placeholder")
+        expect(textarea).toHaveAttribute("name", "test-name")
+        expect(textarea).toHaveValue("test-value")
+    })
+    it("Should be able to type into Textarea", () => {
+        render(<Textarea />)
+        const textareaElement = screen.getByRole<HTMLTextAreaElement>("textbox")
+        fireEvent.change(textareaElement, {target: { value: "testing the textarea element" }})
```
nickytonline, line 33: refactor (blocking): use `userEvent.type` to emulate a user entering text in the text area. This will also fire any other events (if any wired up) along with the onChange event.

You'll also need to install `@testing-library/user-event` and import `userEvent`.

```
pnpm install @testing-library/user-event
```

```typescript
import userEvent from '@testing-library/user-event';
```


```suggestion
    it("Should be able to type into Textarea", async () => {
        render(<Textarea />)
        const textareaElement = screen.getByRole<HTMLTextAreaElement>("textbox")
        await userEvent.type(textareaElement, 'testing the textarea element');
```

Giving you a ping @shashilo as this introduces a new package to the project.

nickytonline (changes requested):
Thanks for the tests @CooleyWC! One change request.

components/TextArea/textarea.test.tsx
CooleyWC: Got it - Should installing the new package be completed in a separate ticket, or can I install/import within this ticket? @shashilo

CooleyWC: @nickytonline @shashilo - fireEvent is now changed to userEvent in the text area test

components/TextArea/textarea.test.tsx
```diff
@@ -0,0 +1,37 @@
...
+// Licensed under the MIT License.
+
+import { render, screen} from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import { Textarea } from './textarea';
+
+describe("Textarea", () => {
+    it("Renders the Textarea correctly", () => {
+        render(<Textarea />)
+        const textarea = screen.getByRole("textbox")
```
shashilo, line 11: Our test style guide is to use `data-testid` to find the elements being tested. Please add this to the component.

CooleyWC: Updated the tests to use data-testid instead of getByRole.

components/TextArea/textarea.test.tsx
```diff
@@ -0,0 +1,37 @@
...
+        render(<Textarea />)
+        const textarea = screen.getByTestId("textarea")
+        expect(textarea).toBeInTheDocument()
+    })
+    it("Renders the Textarea with custom className", ()=>{
+        render(<Textarea className="custom-cn" />)
+        const textarea = screen.getByTestId("textarea")
+        expect(textarea).toHaveClass("custom-cn")
+    })
+    it("Renders the Textarea with props", ()=>{
```
shashilo, line 19: nit: with props, but what props?

shashilo (approved):

shashilo:
Change the PR title to a conventional commit.

components/TextArea/textarea.test.tsx
CooleyWC: That test now reads as: "Renders the Textarea with HTML Attribute props".

shashilo (approved):

kepsteen (approved):
LGTM

nickytonline (approved):
🚢

PR #387, opened 2025-05-04
Fix: closes #251 addresses eslint errors calendar.tsx

### LINKED ISSUE #251: Address ESlint errors in the file: ./components/Calendar/calendar.tsx
## Replicating errors:
- Open terminal
- execute terminal command `npx eslint ./components/Calendar/calendar.tsx`

## Existing errors:
- Line 1:1  
  - Error: missing header  header/header
- Line 12:1  
  - Error: Missing JSDoc comment.  jsdoc/require-jsdoc
- Line 12:1 
  -  Error: Missing return type on function.  @typescript-eslint/explicit-function-return-type
- LIne 12:1
  -   Error: Function component is not an arrow function  react/function-component-definition
- Line 13:3
  -   Error: 'className' is missing in props validation  react/prop-types
- Line 14:3
  -   Error: 'classNames' is missing in props validation  react/prop-types
- Line 15:3 
  -  Error: 'showOutsideDays' is missing in props validation  react/prop-types
- Line 41:17 
  -  Error: 'mode' is missing in props validation  react/prop-types
- Line 66:19  
  - Error: Missing JSDoc comment.  jsdoc/require-jsdoc
- Line 67:20  
  - Error: Missing JSDoc comment.  jsdoc/require-jsdoc

haydnkerr:
fix #251 

This one took a lot more time than I wanted it to. Specifically the missing props validation.

Address following issues:
- Added Header to top of file

Missing props validation :
- Listed validation for each prop in lines 14-18

Missing JSDocs:
- Added JSDocs for Calendar
_ JSDoc for each chevron icons

Missing return type and arrow missing in function
- Converted function into an arrow function


suzuhe481:
Just noticed I accidentally clicked Update Branch on this PR thinking it was mine. It shouldn't have affected anything, but just commenting to let everyone know.

components/Calendar/calendar.tsx
```diff
@@ -1,20 +1,37 @@
...
 import * as React from 'react';
 import { ChevronLeft, ChevronRight } from 'lucide-react';
 import { DayPicker } from 'react-day-picker';
+import { JSX } from 'react';
 
 import { cn } from '@/lib/utils';
 import { buttonVariants } from '@/components/Button/button';
 
-export type CalendarProps = React.ComponentProps<typeof DayPicker>;
+export type CalendarProps = React.ComponentProps<typeof DayPicker> & {
```
kepsteen, line 14: Why are you adding these extra values to the CalendarProps type?

components/Calendar/calendar.tsx
```diff
@@ -63,13 +80,23 @@ function Calendar({
         // The lack of props is causing a linter error, but we will need them when setting up the fetch logic later. For now I'm commenting out the error as a temporary fix.
         // IconLeft: ({ ...props }) => <ChevronLeft className="h-4 w-4" />,
```
kepsteen, line 81: If you uncomment this, is it still causing a linter error?

kepsteen (changes requested):

components/Calendar/calendar.tsx
haydnkerr: Yes it is saying 'props' defined but never used, so i just left it commented out.

components/Calendar/calendar.tsx
```diff
@@ -1,20 +1,32 @@
+// Copyright (c) Gridiron Survivor.
+// Licensed under the MIT License.
+
 'use client';
 
 import * as React from 'react';
```
kepsteen, line 6: Only import what you need from react.  Like how you used `import { JSX } from 'react'` below

components/Calendar/calendar.tsx
kepsteen: Can you try passing the props to the `<ChevronLeft />` component? The comment tends to suggest that we will need to do that for this component to work correctly.

kepsteen (changes requested):

components/Calendar/calendar.tsx
haydnkerr: I've been trying to work on this. Should I just change it to "import React from 'react';

since React.componentProps is initialized later in the code?

kepsteen: What I mean by this is that you're importing all of React into this file, but you're only using `React.JSX` and `React.componentProps`. You can use named imports to import only the modules you need from React. You're already doing this for the JSX import, so you just need to use the same pattern for `componentProps`.

kepsteen (approved):
LGTM

bethanyann (approved):
Read through the other suggested changes and agree with the updates made, therefore approving! Looks good.

PR #395, opened 2025-05-09
Hadi/211-add-unit-tests-for-slider-component

### LINKED ISSUE #211: Create unit test for slider.tsx

hdkassamali:
Closes #211 

### Before:
- No unit tests for `slider.tsx`

### After:
- unit test for rendering the slider component
- unit test for rendering the slider component with a custom class
- unit test for forwarding child props to the slider component
- unit test for mocking a value change when clicking on the slider and using the ArrowRight key

### Note:
- I added a dummy ResizeObserver in the test setup because JSDOM doesn’t implement it, satisfying Radix’s use-size hook (which calls new ResizeObserver() internally) and preventing mount errors in Slider tests.
- I also had to stub out the hasPointerCapture/setPointerCapture/releasePointerCapture methods on HTMLElement.prototype so Radix’s internal pointer-capture calls wouldn't automatically fail in the test environment.

Screenshot 2025-05-09 at 12 28 10 PM


components/Slider/Slider.test.tsx
```diff
@@ -0,0 +1,53 @@
...
+    expect(slider).toHaveClass('my-custom');
+  });
+
+  it('forwards child props to the slider', () => {
+    render(<Slider aria-label="test-aria-label" />);
+    const slider = screen.getByTestId('slider');
+    expect(slider).toHaveAttribute('aria-label', 'test-aria-label');
+  });
+
+  it('calls onValueChange when ArrowRight key is pressed', async () => {
```
kepsteen, line 40: Can you also test that the component respects the min and max boundaries? Otherwise, looks good to me

kepsteen (changes requested):

components/Slider/Slider.test.tsx
hdkassamali: Just added those tests

kepsteen (dismissed):
LGTM

components/Slider/Slider.test.tsx
```diff
@@ -0,0 +1,91 @@
...
+      />,
+    );
+
+    const thumb = screen.getByRole('slider');
+    await user.click(thumb);
+    await user.keyboard('{ArrowRight}');
+    expect(onValueChange).not.toHaveBeenCalled();
+  });
+
+  it("does not decrement below the 'min' boundary", async () => {
```
shashilo, line 73: Need one to decrement the number as expected.

components/Slider/Slider.test.tsx
```diff
@@ -0,0 +1,91 @@
...
+        defaultValue={[0]}
+        min={0}
+        step={10}
+        onValueChange={onValueChange}
+      />,
+    );
+
+    const thumb = screen.getByRole('slider');
+    await user.click(thumb);
+    await user.keyboard('{ArrowLeft}');
```
shashilo, line 88: These are all keyboard events. Is there a way to check by moving the slider like a mouse down event?

components/Slider/Slider.test.tsx
```diff
@@ -0,0 +1,91 @@
+// Copyright (c) Gridiron Survivor.
+// Licensed under the MIT License.
+
+class ResizeObserver {
+  observe() {}
+  unobserve() {}
+  disconnect() {}
+}
+(global as any).ResizeObserver = ResizeObserver;
```
shashilo, line 9: Why do we need to set these global properties?

components/Slider/Slider.test.tsx
```diff
@@ -0,0 +1,91 @@
...
+    expect(slider).toBeInTheDocument();
+  });
+
+  it('renders the component with a custom class name', () => {
+    render(<Slider className="my-custom" />);
+    const slider = screen.getByTestId('slider');
+    expect(slider).toHaveClass('my-custom');
+  });
+
+  it('forwards child props to the slider', () => {
```
shashilo, line 34: I would use the word passes or something else. Forward is a reserved method name.

shashilo (changes requested):

components/Slider/Slider.test.tsx
hdkassamali: I had to stub those globals because Jest’s JSDOM environment doesn’t implement the browser’s ResizeObserver or pointer‐capture APIs that Radix UI uses, so without these mocks the tests would error. Just another weird radix ui thing lol.

components/Slider/Slider.test.tsx
hdkassamali: Adding that to my new PR. Seems like I have to stub the slider width to give the slider a real width so that pointer events can actually move with a mouse event.

kepsteen (approved):
LGTM

components/Slider/Slider.test.tsx
shashilo: I would add it here. You're touching the files and adding test already.

hdkassamali: Shoot sorry, I meant I added that to my new commit*. So the changes for the keyboard events are in here already.

shashilo (approved):

PR #401, opened 2025-05-12
Julianna/190 global gift spinner unit tests

### LINKED ISSUE #190: Create unit test for GlobalGiftSpinner.tsx

mathematiCode:
Closes #190 
I added the following 5 tests for the GlobalGiftSpinner component. 
image
It was tricky for me to understand how this animation was working so I put it in a [code sandbox](https://codesandbox.io/p/sandbox/framer-svg-animation-t2rd4r). I discovered that although it appears to be changing the pathLength, framer motion actually changes the stroke-dasharray instead and keeps the pathLength constant. Since that's not very obvious and hard to get across through my tests themselves, I added a comment to my test file explaining this. 

I wasn't entirely sure of the best way to approach getting the times to test. My choices were either:
- Use 1.5 seconds in the test, and it could break if someone changes the duration of the animation.
- Get the duration that's defined inside of the component to use in my tests. (testing implementation details)
- Grab the values from a few different time increments and make the test pass if _any_ of them are between 0 and 1. (possibly over complicated) 

I went with the first option, but please let me know if you think there's a better way.


lillian-nguyen (approved):
LGTM

bethanyann (approved):
I don't think I'd personally test styling like this because I don't think if the tests fail, they provide useful feedback for what exactly is wrong or causing the failure, but I think this is a decent approach given that this is a framer animation and it could potentially get a little weird with rendering or spinning only once. 

All that to say, approved :)

lillian-nguyen (approved):

PR #402, opened 2025-05-12
[#152] Test: created integration test for supabase/middleware

### LINKED ISSUE #152: Create integration test for supabase/middleware.ts

alexappleget:
Closes #152 

All tests pass and I double checked by forcing them to fail to make sure it was testing correctly.

And since I touched the middleware file to move it into its own folder and write a test for it, I handled the eslint errors for the `middleware.ts` as well.

IMPORTANT: I edited the `middleware.ts` file because it was needed to even make any of these tests pass and be accurate.


lib/supabase/middleware/middleware.ts
```diff
@@ -29,15 +47,10 @@ export async function updateSession(request: NextRequest) {
     },
   );
 
-  // IMPORTANT: Avoid writing any logic between createServerClient and
```
mathematiCode, line 32: Why is this comment being deleted? It seems important.

lib/supabase/middleware/middleware.ts
```diff
@@ -74,18 +82,5 @@ export async function updateSession(request: NextRequest) {
     }
   }
```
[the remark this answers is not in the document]
mathematiCode: I'm going to close this since I created a separate ticket for it here: #405

lib/supabase/middleware/middleware.test.ts
```diff
@@ -0,0 +1,102 @@
...
+
+    expect(NextResponse.redirect).toHaveBeenCalledWith(redirectUrl);
+  });
+
+  it('redirects user to /onboarding if their onboarding is set as false', async () => {
+    const mockUser = { id: '123' };
+
+    (createServerClient as jest.Mock).mockReturnValue({
+      auth: {
+        getUser: jest.fn().mockResolvedValue({ data: { user: mockUser } }),
```
mathematiCode, line 56: I think we should create a test database. One user with onboarding=true and one with onboarding=false so we can call it and test that they can only access the pages they're supposed to.

mathematiCode (changes requested):
I think it's fine to mock the page redirects, but I do think we should have an actual test database with test users to make sure they can only access the appropriate pages. Otherwise, really nice job on this!

mathematiCode (commented):
I see this same logic in a few of the tests just with the email and the request url changing. 
```
  const {
      data: { session },
    } = await supabase.auth.signInWithPassword({
      email: 'onboardingfalse@test.com',
      password: 'password',
    });

    const cookieHeader = `sb-access-token=${session?.access_token}; sb-refresh-token=${session?.refresh_token}`;

    const request = new NextRequest('https://localhost:4000/dashboard', {
      headers: new Headers({
        cookie: cookieHeader,
      }),
    });
```
What do you think of pulling this out into its own helper function that takes in the email and url as parameters? 
Then the rest of us could even use your same function for our integration tests as well bc it seems like we'd need to use this same functionality any time we need to login as a test user.

lib/supabase/middleware/middleware.test.ts
```diff
@@ -1,102 +1,100 @@
...
-      from: () => ({
-        select: () => ({
-          eq: () => ({
-            single: jest.fn().mockResolvedValue({
-              data: { onboarding_complete: false },
-            }),
-          }),
-        }),
-      }),
+  it('allows user to /dashboard if their onboarding is set as true', async () => {
```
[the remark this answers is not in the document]
alexappleget: fixed. i like this naming better you suggested

lib/supabase/middleware/middleware.test.ts
```diff
@@ -0,0 +1,100 @@
...
+import { createClient } from '@supabase/supabase-js';
+
+const supabase = createClient(
+  process.env.NEXT_PUBLIC_SUPABASE_URL!,
+  process.env.NEXT_PUBLIC_SUPABASE_ANON_KEY!,
+);
+
+describe('Middleware', () => {
+  it('redirects unauthenticated user back to /', async () => {
+    const request = new NextRequest('https://localhost:4000/dashboard');
```
shashilo, line 16: We're testing local or staging?

shashilo (changes requested):

lib/supabase/middleware/middleware.test.ts
nickytonline: also make the base URL an environment variable and then you can do something like this.


```suggestion
    const request = new NextRequest(`${new URL('/dashboard', BASE_APP_URL)}`);
```

lib/supabase/middleware/middleware.ts
```diff
@@ -33,11 +51,22 @@ export async function updateSession(request: NextRequest) {
   // supabase.auth.getUser(). A simple mistake could make it very hard to debug
   // issues with users being randomly logged out.
 
-  const {
-    data: { user },
-  } = await supabase.auth.getUser();
+  const accessToken = request.cookies.get('sb-access-token')?.value;
```
nickytonline, line 54: From what I’ve read, `@supabase/ssr` doesn’t include the `createMiddlewareClient` helper like the older `@supabase/auth-helpers-nextjs` package did. So manually reading the cookie to get the access token is the correct approach here.

That said, does this mean middleware-based auth hasn’t been working since we switched to `@supabase/ssr`? Just wondering if we had a gap in session handling before this change.

alexappleget: auth was working for google auth, but i believe we need to clean up our auth context as well. it's not handling things correctly

lib/supabase/middleware/middleware.test.ts
alexappleget: I can't make this change yet as the staging/production url has not been set up yet on vercel

PR #407, opened 2025-05-14
fix #281: address ESLint issues in Skeleton component

### LINKED ISSUE #281: Address ESlint errors in the file: ./components/Skeleton/Skeleton.tsx
Replicating errors:
- Open terminal
- execute terminal command `npx eslint ./components/Skeleton/Skeleton.tsx`

## Existing errors:
- Line 1:1  
  - Error: missing header  header/header
- Line 5:8
  -   Error: Missing JSDoc comment.  jsdoc/require-jsdoc
- Line 5:8 
  -  Error: Missing return type on function.  @typescript-eslint/explicit-function-return-type
- Line 5:8
    -  Error: Function component is not an arrow function  react/function-component-definition

lillian-nguyen:
Address ESLint errors within the Skeleton component by:

- Adding missing 
  - `header/header`
  - JSDoc comment
- Specifying return type on function
- Refactoring component to use arrow function

Additional changes: 
- Added "I" prefix to SkeletonProps to denote interface

closes #281


mathematiCode (approved):
LGTM 

The only thing I'm not sure about is if we need to have an interface for just the one prop, but I can see that was already there before your changes. And I know sometimes teams like to have an interface for every component.

components/Skeleton/Skeleton.tsx
```diff
@@ -1,9 +1,20 @@
-interface SkeletonProps {
+// Copyright (c) Gridiron Survivor.
+// Licensed under the MIT License.
+
+import { JSX } from "react";
+
+interface ISkeletonProps {
```
nickytonline, line 6: refactor (potentially blocking): @shashilo, are we using interface names prefixed with `I`?

```suggestion
interface SkeletonProps {
```

lillian-nguyen: @nickytonline I was told interfaces are prefaced with `I`

nickytonline (approved):
🚢

PR #408, opened 2025-05-14
zoobia/187-unit-test-GiftDetailsView

zoobiasyyed:
completed majority tests, just having trouble checking if the Gift Icon displays if conditional fails

### DIFF
--- a/components/GiftDetailsView/GiftDetailsView.test.tsx
+++ b/components/GiftDetailsView/GiftDetailsView.test.tsx
@@ -1,0 +1,113 @@
+// Copyright (c) Gridiron Survivor.
+// Licensed under the MIT License.
+
+import React from 'react';
+import { render, screen } from '@testing-library/react';
+import GiftDetailsView from './GiftDetailsView';
+import { describe } from 'node:test';
+import { userEvent } from '@storybook/test';
+
+const mockGift = {
+  id: '1',
+  title: 'Blanket',
+  price: '$29.99',
+  description: 'fuzzy blanket',
+  matchReasons: ['fits theme', 'highly rated'],
+  matchScore: 0.98,
+  imageUrl:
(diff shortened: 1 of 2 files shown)


components/GiftDetailsView/GiftDetailsView.test.tsx
```diff
@@ -0,0 +1,113 @@
...
+      <GiftDetailsView gift={mockGift} handleFeedback={mockHandleFeedback} />,
+    );
+
+    const image = screen.getByTestId('valid-image');
+    expect(image).toBeInTheDocument();
+  })
+
+  //testing that Gift Icon appears if not valid image URL
+  //CAN NOT FIGURE THIS OUT
+  it('renders Gift Icon if URl is not valid', ()=> {
```
nickytonline, line 70: The reason this test fails at the moment is because of this line in the GiftDetailsView component, https://github.com/LetsGetTechnical/elecretanta/blob/develop/components/GiftDetailsView/GiftDetailsView.tsx#L34

```
{gift.imageUrl && !imageError ? (
```

As long as `gift.imageUrl` has a string value, this check assumes the gift image URL is valid. Check to see if it's a valid URL. Your mock bad gift has a valid string for a url which is why your conditional never works like you asked in Discord @zoobiasyyed.

```typescript
const mockBadGift = {
  id: '1',
  title: 'Blanket',
  price: '$29.99',
  description: 'fuzzy blanket',
  matchReasons: ['fits theme', 'highly rated'],
  matchScore: 0.98,
  imageUrl:'bad_url',
}
```

You can use a helper function like this:

```diff
+ const isValidUrl = (urlString: string) => {
+   try {
+     new URL(urlString);
+     return true;
+   } catch {
+     return false;
+   }
+ };

- {gift.imageUrl && !imageError ? (
+ {gift.imageUrl && isValidUrl(gift.imageUrl) && !imageError ? ( 
```

nickytonline (changes requested):

components/GiftDetailsView/GiftDetailsView.test.tsx
zoobiasyyed: thank you so much @nickytonline gonna try this out

components/GiftDetailsView/GiftDetailsView.tsx
```diff
@@ -10,6 +10,15 @@ import { SquareArrowOutUpRight, ThumbsDown, Gift } from 'lucide-react';
 import { GiftSuggestion } from '@/app/types/giftSuggestion';
 import { useState, useCallback } from 'react';
 
+const isValidUrl = (urlString: string) => {
```
nickytonline, line 13: This works great! It makes sense to keep it in here for now as nothing else is using it at the moment. If something else needs this in the future, we can pull it out of this component and promote it to a utility function and at that point we can also add unit tests for `isValidUrl`.

nickytonline (approved):
Thanks for making the changes @zoobiasyyed! :shipit:

nickytonline:
Looks like you just need to update the branch @zoobiasyyed.

bethanyann (approved):
Approving since Nick's change requests were addressed

PR #427, opened 2025-05-20
test #215: created unit test for WarningModal component

alexappleget:
Closes: #215 

NOTE!!: I didn't put the `render()` inside a beforeEach for the `WarningModal` component because the auth context has to be handled first and it is different for each test

image


bethanyann (dismissed):
Still kinda new to reading jest tests, but read through the different assertions and they all make sense as far as what you're testing. Appreciated your note about rendering the component in each test because of the different authContext needed.

hdkassamali (approved):
LGTM

bethanyann (approved):
Nice work!

PR #445, opened 2025-05-22
test #154: created unit and integration tests for SnowOverlayProvider

### LINKED ISSUE #154: Create unit test for SnowOverlayProvider.tsx

suzuhe481:
Closes #154

## Unit Tests
- Checks SnowOverlayProvider's isSnowing default value.
- Checks that SnowOverlayProvider's toggleSnowSetting works as intended.

## Integration Tests
Created a component which will be wrapped by the SnowOverlayProvider and use its values.
- Reads and updates localStorage through SnowOverlayProvider's toggle.
- Reads an already stored value in localStorage.

## Possible bugs in SnowOverlayProvider.tsx
**Not all tests currently pass.** I believe this is due to how ```SnowOverlayProvider.tsx``` stores default values in state and localStorage. I have my suggested changes below.

- SnowOverlayProvider currently stores isSnowing as a state with a default value of true. Since it's stored in localStorage, we should check if it exists in localStorage before setting the state.
```
// old
const [isSnowing, setIsSnowing] = useState(true);

// suggestion
 const [isSnowing, setIsSnowing] = useState(() => {
  const storedIsSnowing = localStorage.getItem('isSnowing');
  return storedIsSnowing === null ? true : JSON.parse(storedIsSnowing);
});
```

- There is a useEffect in SnowOverlayProvider.tsx which sets the isSnowing state on the initial page load. I don't believe it's doing it correctly. It should get the value from localStorage and if it doesn't exist, set it and the state to the default value of true.

```
// old
useEffect(() => {
    if (typeof window !== 'undefined') {
      const setting = localStorage.getItem('isSnowing');
      if (setting !== null) {
        setIsSnowing(JSON.parse(setting));
      }
    }
}, []);

// suggestion
useEffect(() => {
    const storedIsSnowing = localStorage.getItem('isSnowing');

    if (storedIsSnowing === null) {
      localStorage.setItem('isSnowing', 'true');
      setIsSnowing(true);
    }
}, []);
```

- The useEffect can also be completely removed and combined with the useState function with this.
```
const [isSnowing, setIsSnowing] = useState(() => {
    const storedIsSnowing = localStorage.getItem('isSnowing');

    if (storedIsSnowing === null) {
      localStorage.setItem('isSnowing', 'true');
      return true;
    }

    return JSON.parse(storedIsSnowing);
});
```


lillian-nguyen (changes requested):
suggest adding a check to verify localStorage is updated after toggling. Something like `expect(localStorage.getItem('isSnowing)).toBe('false');`

providers/SnowOverlayProvider.test.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+    const snowToggleButton = screen.getByTestId('snow-button');
+
+    await waitFor(() => {
+      expect(snowToggleButton).toHaveTextContent('Snowing');
+    });
+
+    fireEvent.click(snowToggleButton);
+
+    expect(snowToggleButton).toHaveTextContent('Not Snowing');
```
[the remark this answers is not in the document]
suzuhe481: Thanks. I added new checks to check for localStorage values.
Though in doing that, I found that SnowOverlayProvider.tsx might have some bugs in how it sets and stores its states. So while some of my tests fail, I think it's failing "correctly" and will pass when the other bugs are fixed.

providers/SnowOverlayProvider.test.tsx
```diff
@@ -0,0 +1,82 @@
...
+    const { result } = renderHook(() => useSnowOverlay(), {
+      wrapper: SnowOverlayProvider,
+    });
+
+    act(() => {
+      result.current.toggleSnowSetting();
+    });
+
+    expect(result.current.isSnowing).toBe(false);
+    expect(localStorage.getItem('isSnowing')).toBe('false');
```
nickytonline, line 29: refactor (blocking): no need to check values in local storage. The `useLocalStorage` hook is an implementation detail of the provider. Checking `result.current.isSnowing` is all that is required.

Although local storage is an implementation detail, we will need to clear it before each test like the test file is already doing.


```suggestion
    expect(result.current.isSnowing).toBe(false);
```

Same goes for the other tests.

nickytonline (changes requested):

providers/SnowOverlayProvider.test.tsx
suzuhe481: I understand your reason for not needing to check the values in local storage, since testing for `result.current.isSnowing` should be enough. But testing for it is how I found the bug of localStorage not being set properly on the initial page load.

Is it still ok to not test for localStorage? Or would that be unnecessary once the useLocalStorage hook is used like you said?

nickytonline: We have the useLocalStorage hook now and that's what should test if it's actually set in local storage. For this test it doesn't make sense because you're testing the implementation of something that is used in the provider. If we hadn't moved to the useLocalStorage hook, then testing if it was set in local storage here would have made sense potentially.

suzuhe481: Removed the assertions for localStorage.

nickytonline (approved):
🚢

PR #485, opened 2025-05-30
test #206: created unit test for ProfileCard Component

### LINKED ISSUE #206: Create unit test for ProfileCard.tsx

CooleyWC:
ProfileCard Unit test:

Closes #206

• Mocked Radix Avatar component 
    • Verifies the testProfile.avatar is passed as the src to AvatarImage.
    • If avatar is undefined, confirms that the AvatarFallback renders the default image. 
• Next/navigation router, mocks the push method for the handleEditProfile function.
• Created a testProfile object using the Profile type with fake data.

Test Coverage: 
• Renders the following ProfileCard props: profileName, avatar, age_group, hobbies, and avoid.
• Renders a Badge component for each category.
• Triggers navigation via the editProfileButton.
• Displays preference scale values for practical_whimsical, cozy_adventurous, and minimal_luxurious.


components/ProfileCard/ProfileCard.test.tsx
```diff
@@ -0,0 +1,166 @@
...
+
+  it('Edit Profile Button navigates to the onboarding path after user click', async () => {
+    render(<ProfileCard profile={testProfile} showEditButton={true} />);
+    const editProfileButton = screen.getByTestId('editProfileButton');
+    await userEvent.click(editProfileButton);
+    expect(mockPush).toHaveBeenCalledTimes(1);
+    expect(mockPush).toHaveBeenCalledWith('/onboarding?editing=true');
+  });
+
+  it('Renders a Badge for each category', () => {
```
suzuhe481, line 109: The Badge component has its own test id, so that can be used to get all the badges instead of just getting their parent component.

You can also confirm that the number of badges equals the number of categories in the profile.

components/ProfileCard/ProfileCard.test.tsx
```diff
@@ -0,0 +1,166 @@
...
+    render(<ProfileCard profile={testProfile} showEditButton={true} />);
+    const categoryBadges = screen.getByTestId('categoryBadges');
+    expect(categoryBadges).toHaveTextContent(testProfile.categories[0]);
+    expect(categoryBadges).toHaveTextContent(testProfile.categories[1]);
+    expect(categoryBadges).toHaveTextContent(testProfile.categories[2]);
+  });
+
+  it('Renders the preferenceRight, preferenceLeft labels when value is undefined', () => {
```
suzuhe481, line 117: This only tests for the practical-whimsical value. Should there be more checks for the cozy-adventurous and minimal-luxurious values or is this enough to confirm it works for those properties too?

components/ProfileCard/ProfileCard.test.tsx
CooleyWC: Now uses the badge component test-id. Now confirms the number of badges equals the number of categories in the testProfile. Removed the 'categoryBadges' testid from the parent div in ProfileCard.

components/ProfileCard/ProfileCard.test.tsx
CooleyWC: Added tests for cozy-adventurous and minimal-luxurious values

bethanyann (approved):
Nice work!

PR #489, opened 2025-06-03
DRAFT Julianna/playwright-POC

mathematiCode:
This is just a draft of changes we will need to make to set up playwright.


nickytonline:
You can set the PR to draft instead of having it ready for review with the word DRAFT in the title.

CleanShot 2025-06-03 at 23 37 00@2x

Also consider a more descriptive title than `DRAFT Julianna/playwright-research`, e.g. `Playwright POC`

nickytonline:
One thing to consider for end to end tests is it's running against a real application with a real backend. Tests you run can affect subsequent tests. We'll need to look into being able to rollback the database after each test or go back to a snapshot of the database.

package.json
```diff
@@ -33,6 +33,7 @@
     "cmdk": "^1.0.0",
     "cross-env": "^7.0.3",
     "date-fns": "^3.6.0",
+    "dotenv": "^16.5.0",
```
shashilo, line 36: Are the environment variables being setup on the server?

PR #491, opened 2025-06-10
docs #433 created story for LinkCustom component

### LINKED ISSUE #433: Create a Story for the LinkCustom Component
### [Atomic Design](https://atomicdesign.bradfrost.com/chapter-2/) Overview:

Atomic design is a methodology for organizing and scaling components using the following hierarchical building blocks: atoms, molecules, organisms, templates, and pages. It offers a structured approach that considers these "blocks" both in isolation and in combination with one another, helping designers and developers build scalable, consistent user interfaces. 

Atom: The smallest functional block of the UI, like a button, input, or label. These are not broken down further and act as the foundation for more complex components.

Molecule: 2+ atoms functioning together as a unit. For example, a form field with a label, input, and error message combines multiple atoms into a reusable component.

Organism: Relatively complex UI components made up of groups of molecules and/or atoms. An example could be a header containing a logo, navigation, and search form.

Template: Page-level object that arrange components in a way that illustrates content structure (without real content) and layout patterns. An example would be a homepage template. 

Page: Specific instance of a template populated with real content and data. These represent the final, rendered view users interact with. 

### Atomic design type: Atom

- [ ] Create a Story with all separate variants along with component and variant descriptions ([reference](https://react.carbondesignsystem.com/))
- [ ] Create a playground
- [ ] Write documentation to clarify the component ([reference](https://circuit.sumup.com/?path=/story/components-button--variants&args=variant:tertiary))

mathematiCode:
Closes #433 
I created this story for the LinkCustom component
I made the background for this story dark mode so the white text would show up. 
image

### DIFF
--- a/app/api/getUserAvatar/route.ts
+++ b/app/api/getUserAvatar/route.ts
@@ -1,0 +1,25 @@
+// Copyright (c) Gridiron Survivor.
+// Licensed under the MIT License.
+
+import { NextResponse } from 'next/server';
+import { createClient } from '../../../lib/supabase/server';
+
+/**
+ * Get user avatar URL
+ * @returns {Promise<NextResponse>} Promise that resolved to a Response object
+ */
+export async function GET(): Promise<NextResponse> {
+  const supabase = await createClient();
+  const {
+    data: { user },
+    error: userError,
+  } = await supabase.auth.getUser();
+
+  if (userError) {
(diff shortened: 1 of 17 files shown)


suzuhe481 (approved):
LGTM!

I was experimenting a bit to see if it's possible to get tailwind styles to update in real time when changing the className in the story. But I think it was definitely more effort than it's worth. It was just me being curious.
This is already more than enough. Good work

PR #497, opened 2025-06-18
feat: implement useLocalStorage hook and refactor SnowOverlayProvider

### LINKED ISSUE #496: Provider sets its value from both useState and localStorage
### Component
The `/providers/SnowOverlayProvider.tsx` file provides two values for the app.
 - isSnowing - boolean
 - toggleSnowSetting - function

### Issue 
`isSnowing` is a state variable that has it's initial value hard coded to`true` in the useState initializer, but it also gets set by localStorage inside a useEffect where it retrieves the value from localStorage, then sets it to state. This causes a sync problem where the state and localStorage can have different values.

### Proposal
Remove the useEffect, and set the `isSnowing` state based on what's available in localStorage.

Image

suzuhe481:
### Before: 
No useLocalStorage hook.

### After: 
Creates a custom useLocalStorage hook and implements it in `/providers/SnowOverlayProvider.tsx`.

### Summary: 
The `SnowOverlayProvider.tsx` file syncs a value between useState and localStorage, but it would sometimes have mismatched values since the useState was hardcoded. 

I created a custom localStorage hook to fix the error and also simplify localStorage usage throughout the app.

Closes #496
 
## Additional information
Usage works similar to a useState hook. Takes in the type of the value being stored. Then returns a tuple with the variable and setter function.

```tsx
import useLocalStorage from '@/hooks/useLocalStorage';

// Use when reading and writing to localStorage.
const [theme, setTheme] = useLocalStorage<'light' | 'dark'>('theme', 'light');

// Use when only reading from localStorage.
const [theme] = useLocalStorage<'light' | 'dark'>('theme', 'light');
// OR if TypeScript errors occur...
const [theme, _] = useLocalStorage<'light' | 'dark'>('theme', 'light');

```

## Pre-submission checklist

- [ ] Request reviews from the `Peer Code Reviewers` and `Senior+ Code Reviewers` groups
- [ ] Thread has been created and PR is linked in `gis-code-questions`

### DIFF
--- a/hooks/useLocalStorage.ts
+++ b/hooks/useLocalStorage.ts
@@ -1,0 +1,48 @@
+// Copyright (c) Gridiron Survivor.
+// Licensed under the MIT License.
+
+import { useState, useEffect } from 'react';
+import { Dispatch, SetStateAction } from 'react';
+
+/**
+ * Custom useLocalStorage hook.
+ * @template T - Type of the value being stored.
+ * @param key {string} - The localStorage key.
+ * @param initialValue {T} - The initial value to set if none is found in localStorage.
+ * @returns {[T, Dispatch<SetStateAction<T>>]} - The current value and a function to set the value
+ */
+const useLocalStorage = <T>(
+  key: string,
+  initialValue: T,
+): [T, Dispatch<SetStateAction<T>>] => {
+  const isServer = typeof window === 'undefined';
+
+  const [storedValue, setStoredValue] = useState<T>(() => {
+    if (isServer) {
+      return initialValue;
+    }
+
+    try {
+      const item = window.localStorage.getItem(key);
+      return item ? JSON.parse(item) : initialValue;
+    } catch {
+      return initialValue;
+    }
+  });
+
+  useEffect(() => {
+    if (isServer) {
+      return;
+    }
+
+    try {
+      window.localStorage.setItem(key, JSON.stringify(storedValue));
+    } catch {
+      // Catches error
+    }
+  }, [key, storedValue, isServer]);
+
+  return [storedValue, setStoredValue] as const;
+};
+
+export default useLocalStorage;
--- a/providers/SnowOverlayProvider.tsx
+++ b/providers/SnowOverlayProvider.tsx
@@ -1,6 +1,7 @@
 'use client';
 
-import { createContext, useContext, useEffect, useState } from 'react';
+import { createContext, useContext } from 'react';
+import useLocalStorage from '@/hooks/useLocalStorage';
 
 export const SnowOverlayContext = createContext({
   isSnowing: false,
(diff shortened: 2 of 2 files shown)


hooks/useLocalStorage.ts
```diff
@@ -0,0 +1,48 @@
...
+  const isServer = typeof window === 'undefined';
+
+  const [storedValue, setStoredValue] = useState<T>(() => {
+    if (isServer) {
+      return initialValue;
+    }
+
+    try {
+      const item = window.localStorage.getItem(key);
+      return item ? JSON.parse(item) : initialValue;
```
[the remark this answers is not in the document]
suzuhe481: Applied your suggestion to better check for a falsy values.

hooks/useLocalStorage.ts
```diff
@@ -0,0 +1,44 @@
...
+    }
+  });
+
+  useEffect(() => {
+    try {
+      window.localStorage.setItem(key, JSON.stringify(storedValue));
+    } catch (error) {
+      throw new Error(`Failed to set local storage item : ${error}`);
+    }
+  }, [key, storedValue, isServer]);
```
nickytonline, line 39: refactor (blocking): `isServer` is no longer required in the dependency array.

```suggestion
  }, [key, storedValue]);
```

hooks/useLocalStorage.ts
```diff
@@ -0,0 +1,44 @@
+// Copyright (c) Gridiron Survivor.
+// Licensed under the MIT License.
+
+import { useState, useEffect } from 'react';
+import { Dispatch, SetStateAction } from 'react';
+
```
nickytonline, line 6: refactor (nit): use a type alias to communicate better what the type is, e.g.

```suggestion
type LocalStorageReturnType<T> = [T, Dispatch<SetStateAction<T>>]

```

and use it everywhere `[T, Dispatch<SetStateAction<T>>]` is.

nickytonline (changes requested):
one more small change request and this is good to go,

hooks/useLocalStorage.ts
suzuhe481: Removed the unused dependency.

nickytonline (approved):
🚢

hooks/useLocalStorage.ts
```diff
@@ -0,0 +1,46 @@
...
+    } catch {
+      return initialValue;
+    }
+  });
+
+  useEffect(() => {
+    try {
+      window.localStorage.setItem(key, JSON.stringify(storedValue));
+    } catch (error) {
+      throw new Error(`Failed to set local storage item : ${error}`);
```
nickytonline, line 39: refactor (block): The throw is inside a `useEffect`, which means any error thrown will be caught by [React's error boundary](https://react.dev/reference/react/Component#catching-rendering-errors-with-an-error-boundary) (if one exists) or bubble up as an unhandled promise rejection. Since `useEffect` doesn't return anything that the calling code can catch, throwing here doesn't provide a meaningful way for the component to handle the error.

Instead just log the error or set an error in state, e.g. useState and surface that.

nickytonline (changes requested):
This is looking good. One last change request.

hooks/useLocalStorage.ts
suzuhe481: I was thinking that and decided it was better to just throw the error to get caught by React since we didn't have anything to catch it.
I was first going to just console log the error, but our eslint rules prevent console logs, so I decided to just throw the Error. 

I'll do your suggestion of storing the error in state. I was trying to get the actual error type and set it's message to the state. But I kept getting type errors, so I just stored a string.

PR #501, opened 2025-06-20
fix: #123 adjusting linting issues

shashilo:


.eslintignore
```diff
@@ -29,6 +29,7 @@ app/types/giftExchange.ts
 app/types/giftExchangeMember.ts
 app/types/profile.ts
 app/types/tailwindcss-motion.d.ts
+components/Avatar/AvatarFallback.tsx
```
nickytonline, line 32: I probably don't have the context, but why is this being added to the eslint ignore file?

nickytonline (approved):

PR #536, opened 2025-07-01
Add #435: Storybook stories for Progress component

haydnkerr:
Before: No Story in place for Progress component as part of issue #435

After: 
- Create Empty, Halfway and Full Progress stories showcasing different progress values
- Includes interactive controls for value, className and indicatorClassName.

Issue: Progress bar renders utilizing a parent component however this comes with an 'error' that I am still figuring out.


stories/Progress.stories.tsx
```diff
@@ -0,0 +1,59 @@
+import type { Meta, StoryObj } from '@storybook/react';
+
+import { Progress } from '../components/Progress/progress';
+import { Indicator } from '@radix-ui/react-progress';
```
suzuhe481, line 4: Don't think you need to import Indicator. I removed it and the story still works.

stories/Progress.stories.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+export default meta;
+
+type Story = StoryObj<typeof meta>;
+
+export const Empty: Story = {
+  args: {
+    value: 0,
+  },
+  render: (args) => (
```
suzuhe481, line 32: Look up decorators in Storybook. This is a way to add a wrapper to the element so you can wrap each Progress in a div.
Then you can remove all the `render`s you have in each export.
[https://storybook.js.org/docs/writing-stories/decorators](https://storybook.js.org/docs/writing-stories/decorators)

If it complains about a missing return type, import and use ReactElement. There might be a better type, but I found this works.

stories/Progress.stories.tsx
```diff
@@ -0,0 +1,59 @@
...
+    <div style={{ width: '300px' }}>
+      <Progress {...args} />
+    </div>
+  ),
+};
+
+export const halfway: Story = {
```
suzuhe481, line 39: Export names should be PascalCase.

suzuhe481:
In the Gridiron Survivor repo, stories were located inside their respective component instead of inside the stories folder. We should get confirmation, but I'm pretty sure we'll keep them in their respective components.

stories/Progress.stories.tsx
```diff
@@ -0,0 +1,61 @@
+import type { Meta, StoryObj } from '@storybook/react';
+
+import { Progress } from '../components/Progress/progress';
+import { Indicator } from '@radix-ui/react-progress';
+
+import * as React from 'react';
+
+const meta: Meta<typeof Progress> = {
```
suzuhe481, line 8: Rewrite the meta to follow the way types are used like in the example stories in the story folder. While this probably works, it's better to follow what's used in the documentation examples. 

```
const meta = { 
...
} satifies Meta<typeof Progress>;
```

stories/Progress.stories.tsx
```diff
@@ -0,0 +1,59 @@
+import type { Meta, StoryObj } from '@storybook/react';
```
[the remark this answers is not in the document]
haydnkerr: Headers have been added

stories/Progress.stories.tsx
haydnkerr: Import has been removed

components/Progress/Progress.stories.tsx
```diff
@@ -0,0 +1,69 @@
...
+      control: 'text',
+      defaultValue: 'relative h-2 w-screen overflow-hidden rounded-full bg-primary/20',
+    },
+    value: {
+      control: { type: 'range', min: 1, max: 100, step: 1 },
+      defaultValue: 25
+    },
+    indicatorClassName: {
+      control: 'text',
+      defaultValue: 'h-full w-screen flex-1 bg-[#E26969] transition-all rounded-full',
```
nickytonline, line 41: refactor (nit): using `w-screen` for width, but the Storybook decorator restricts the parent to `width: 300px`. For a consistent, contained progress bar, use w-full instead of w-screen in your className/indicatorClassName defaults.

nickytonline (dismissed):
🚢

stories/Progress.stories.tsx
suzuhe481: There's still a missing return type on the decorator. Use `import { ReactElement } from 'react';` instead of importing all of React.
Then use ReactElement as the decorator's return type.

stories/Progress.stories.tsx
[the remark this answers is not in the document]
suzuhe481: Just add a newline after the headers to fix the eslint error.

suzuhe481:
The build is failing. Pull from develop branch and run `pnpm i` to keep up to date with develop. Remember to do this regularly.

nickytonline (approved):
🚢

Issue #517, opened 2025-06-24
Storybook configuration


suzuhe481:
Storybook configuration doesn't support Tailwind styles. The globals.css file needs to be imported into Storybook's `/.storybook/preview.ts` file.

Add this line to the top of that file.
`import '../app/globals.css';`

The preview file can also be used for adding more configuration options. These are some options that can be discussed.
- Centering the component on screen.
- Setting multiple background colors that can be changed. Useful for viewing components that are harder to see on certain backgrounds.

PR #541, opened 2025-07-02
docs: #517 add story for Button component

### LINKED ISSUE #430: Create a Story for the Button Component
### [Atomic Design](https://atomicdesign.bradfrost.com/chapter-2/) Overview:

Atomic design is a methodology for organizing and scaling components using the following hierarchical building blocks: atoms, molecules, organisms, templates, and pages. It offers a structured approach that considers these "blocks" both in isolation and in combination with one another, helping designers and developers build scalable, consistent user interfaces. 

Atom: The smallest functional block of the UI, like a button, input, or label. These are not broken down further and act as the foundation for more complex components.

Molecule: 2+ atoms functioning together as a unit. For example, a form field with a label, input, and error message combines multiple atoms into a reusable component.

Organism: Relatively complex UI components made up of groups of molecules and/or atoms. An example could be a header containing a logo, navigation, and search form.

Template: Page-level object that arrange components in a way that illustrates content structure (without real content) and layout patterns. An example would be a homepage template. 

Page: Specific instance of a template populated with real content and data. These represent the final, rendered view users interact with. 

### Atomic design type: Atom

- [ ] Create a Story with all separate variants along with component and variant descriptions ([reference](https://react.carbondesignsystem.com/))
- [ ] Create a playground
- [ ] Write documentation to clarify the component ([reference](https://circuit.sumup.com/?path=/story/components-button--variants&args=variant:tertiary))

suzuhe481:
## Description
Creates a story for the Button component.

### Before: 
No story for Button component.

### After: 
Creates a story for the Button component.

Closes #430 

## Pre-submission checklist

- [ ] Thread has been created in Discord and PR is linked in `gis-code-questions`

### DIFF
--- a/components/Button/Button.stories.ts
+++ b/components/Button/Button.stories.ts
@@ -1,0 +1,95 @@
+// Copyright (c) Gridiron Survivor.
+// Licensed under the MIT License.
+
+import type { Meta, StoryObj } from '@storybook/react';
+import { Button } from '@/components/Button/button';
+
+const meta = {
+  title: 'Button',
+  component: Button,
+  parameters: {
+    docs: {
+      description: {
+        component: `A customizable Button Component.`,
+      },
+    },
+  },
+  tags: ['autodocs'],
+  argTypes: {
+    variant: {
+      description: 'The style of the button',
+    },
+    size: {
+      description: 'The size of the button',
+    },
+    children: {
+      description: 'The content of the button',
+    },
+    asChild: {
+      description: 'Renders as child component if true',
+    },
+  },
+} satisfies Meta<typeof Button>;
+
+export default meta;
+type Story = StoryObj<typeof meta>;
+
+export const Default: Story = {
+  args: {
+    variant: 'default',
+    size: 'default',
+    children: 'Button',
+  },
+};
+
+export const Destructive: Story = {
+  args: {
+    variant: 'destructive',
+    size: 'default',
+    children: 'Button',
+  },
+};
+
+export const Outline: Story = {
+  args: {
+    variant: 'outline',
+    size: 'default',
+    children: 'Button',
+  },
+  parameters: {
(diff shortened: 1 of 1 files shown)


components/Button/Button.stories.ts
```diff
@@ -0,0 +1,95 @@
+// Copyright (c) Gridiron Survivor.
+// Licensed under the MIT License.
+
+import type { Meta, StoryObj } from '@storybook/react';
```
nickytonline, line 4: Refactor (blocking):

Missing Coverage:
 
- Size variations, 'default' is handled but your component supports 'sm', 'lg', and icon sizes
- Disabled state - Important for accessibility testing
- With icons - Your component has specific SVG styling rules
- asChild functionality, key feature that's not demonstrated
- Long text handling, see how the button behaves with longer content

nickytonline (changes requested):

components/Button/Button.stories.ts
suzuhe481: > * Size variations, 'default' is handled but your component supports 'sm', 'lg', and icon sizes

Added options to allow users to set size and variant in the story.

> * Disabled state - Important for accessibility testing

Added a disabled property.
 
> * With icons - Your component has specific SVG styling rules

I made a new story(With Icon) to show this feature, but I had to use .tsx and render() in order to make the story since storybook can't render raw tsx as a string. I also disabled the controls on this story since the values can't interact with the story.

> * asChild functionality, key feature that's not demonstrated

Made a new As Child story, and same situation as the With Icon story.

> * Long text handling, see how the button behaves with longer content

This is the only thing I didn't address since I believe it's already handled with the children property. The user can change the value here which affects the content of the component.

nickytonline (approved):
thanks for making the changes! :shipit:

PR #587, opened 2025-08-26
Fix: fixed warning modal issue when inviting users to a group.

### LINKED ISSUE #582: Opening Group Invite link doesn't open modal on some browsers
When opening a group invite link to join a group, the modal to confirm joining doesn't appear on some browsers.

Works on...
- Chrome

Doesn't work on...
- Firefox

alexappleget:
### Before: 
On the staging environment whenever you would get invited to the gift exchange (but weren't logged in) the warning modal wouldn't show up and throw errors and not load things. 

### After: 
I added the `else { setIsMember(false) } because it'll properly set the state to `false` if the person doesn't have a session. The warning model shows up if this state is `false` and before when the user didn't have a session, it was keeping it as `null`. Adding `null` into the conditional rendering wasn't working previously because it would cause it to flicker the modal before showing it because the initial state is always `null` to start.

 Closes #582 

## Testing instructions

Create a gift exchange group yourself on the staging and/or localhost side and send other members the invite link to have them test if the modal pops up and they can join with no issues.

## Pre-submission checklist

### DIFF
--- a/app/gift-exchanges/[id]/page.tsx
+++ b/app/gift-exchanges/[id]/page.tsx
@@ -79,6 +79,8 @@ export default function GiftExchangePage() {
             (member: GiftExchangeMember) => member.user_id === session?.user.id,
           ),
         );
+      } else {
+        setIsUserAMember(false);
       }
     } catch (error) {
       console.error('Error fetching data:', error);


lillian-nguyen (dismissed):
LGTM

app/api/gift-exchanges/[id]/route.ts
```diff
@@ -19,11 +19,7 @@ export async function GET(
       .single();
 
     if (error) {
-      console.error('supabase error', error);
-      return NextResponse.json(
-        { error: 'Gift exchange not found' },
-        { status: 404 },
-      );
+      throw error;
```
Comment (nickytonline) [inline app/api/gift-exchanges/[id]/route.ts:22]: we throw the error now, but where is it handled?

PR #589, opened 2025-08-27
Fix: set default value to 50 and set min value to 0

### LINKED ISSUE #567: Gift Compass Onboarding Step Doesn't submit on 50%
P0: User should be able to submit at 50% for the range.

P4: The form validation (a good thing) causes content layout shift when validation fails.

Image

alexappleget:
### Before: 
When the slider was more to the left it wasn't submitting because initially the slider's min value was -100 and max value was 100 with 0 being the default value. Anything to the left was a negative value which failed to update supabase because in our database the value has to be `>= 0`.

### After: 
Set min value to 0 and then changed the default starting value to 50.

 Closes #567 

## [optional] Screenshots

https://github.com/user-attachments/assets/5da26cfd-697f-43ae-a6a9-b627d2f2a731

## Pre-submission checklist

### DIFF
--- a/app/onboarding/page.tsx
+++ b/app/onboarding/page.tsx
@@ -176,9 +176,9 @@ function Onboarding() {
       categories: [],
       hobbies: '',
       giftRestrictions: '',
-      giftPersonality: 0,
-      experienceStyle: 0,
-      giftStyle: 0,
+      giftPersonality: 50,
+      experienceStyle: 50,
+      giftStyle: 50,
     },
   });
 
@@ -444,7 +444,7 @@ function Onboarding() {
                           <FormLabel>Gift Personality</FormLabel>
                           <FormControl>
                             <Slider
-                              min={-100}
+                              min={0}
                               max={100}
                               value={[field.value]}
                               onValueChange={(values) =>


nickytonline (commented):
The fix looks good. Are we adding any tests for this or is that a separate issue, or are there no tests for this form yet (probably better as an E2E test which we're working on setting up)?

lillian-nguyen (approved):
LGTM

app/onboarding/page.test.tsx
```diff
@@ -0,0 +1,66 @@
...
+  }),
+}));
+
+describe('Onboarding Page sliders', () => {
+  it('renders all sliders with default 50, min 0, and max 100 values', async () => {
+    render(<OnboardingPage />);
+
+    const nextButton = screen.getByTestId('next-button');
+
+    for (let i = 0; i < 4; i++) {
```
shashilo, line 45: Instead of stepping through this, just change currentStep to 4.

app/onboarding/page.test.tsx
```diff
@@ -0,0 +1,66 @@
...
+    const giftStyleSlider = await screen.findByTestId('gift-style');
+
+    expect(giftPersonalitySlider).toBeInTheDocument();
+    expect(experienceStyleSlider).toBeInTheDocument();
+    expect(giftStyleSlider).toBeInTheDocument();
+
+    const sliders = screen.getAllByRole('slider');
+    expect(sliders.length).toBe(3);
+
+    sliders.forEach((slider) => {
```
shashilo, line 60: If one of these sliders fails, how will we know which one failed?

shashilo (changes requested):

shashilo (approved):

Issue #570, opened 2025-08-05
Profile picture in upper right does not update


BaconDesigns:
Once user is logged in with Google, the picture on the Dashboard/upper right does not update. Stays on the default image.

<img width="240" height="233" alt="Image" src="https://github.com/user-attachments/assets/bd633dc3-1a52-4767-a698-a945576845f5" />

shashilo:
Talk to @lillian-nguyen about this work.

PR #603, opened 2025-09-11
Bug #570: avatar is now loading correctly. Was passing an object through the parameters and not just the avatar url string.

alexappleget:
### Before: 
Google avatar was not displaying in the top right global header bar.

### After: 
Avatar image is now showing up. The issue was that it originally being passed through as: `{avatarUrl: "exampleAvatar.url"}` but now passes through as: `"exampleAvatar.url"`.

 Closes #[ticketnumber]

## Testing instructions

Try out with your own avatar by logging in.
 
## Additional information

The shadcn avatar component only accepts strings, but it was receiving an `object`. Typescript didn't catch it because there was no `Promise<string>` on the `getUserAvatar` function. When typescript sees response, it infers it as `any` and doesn't throw errors unless the response function promises a certain return.

## Pre-submission checklist

### DIFF
--- a/app/assets/svgs/LogoCombination.tsx
+++ b/app/assets/svgs/LogoCombination.tsx
@@ -139,6 +139,15 @@ const LogoCombination: React.FC<React.HTMLAttributes<HTMLDivElement>> = ({
         fill="#FAB76A"
         id="path24"
       />
+      <text
+        x="340"
+        y="55"
+        fill="#ffffff"
+        fontSize="16"
+        fontFamily="Arial, sans-serif"
+      >
+        BETA
+      </text>
     </svg>
   );
 };
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "elecretanta",
-  "version": "0.7.0-alpha",
+  "version": "0.8.0-alpha",
   "private": true,
   "scripts": {
     "dev": "cross-env PORT=4000 next dev",


lib/getUserAvatar.ts
```diff
@@ -13,7 +13,7 @@ const getUserAvatar = async (): Promise<string> => {
     headers: { 'Content-Type': 'application/json' },
   });
   const userAvatar = await response.json();
-  return userAvatar;
+  return userAvatar.avatarUrl;
```
nickytonline: What else is on the user avatar object? Is any of that other info useful for the app? If so, I'd keep it as returning the full avatar object and do `userAvatar.avatarUrl` where the url is needed.

Also, if it's an avatar object, no need to call the URL `avatarUrl`, but that could be something else to visit another day.

alexappleget: the api only returns the avatar's url via:
```tsx
const avatarUrl = user?.user_metadata.avatar_url;

  return NextResponse.json({ avatarUrl });
 ```

The api returns only the person's google avatar url, but whoever made this api made it return as an object that isn't necessary. I can just leave the frontend code as it was with `return userAvatar` and then just delete the `{}` from that api response.

Before it returned just `{userAvatar: 'url'}` instead of just `userAvatar: 'url'`

nickytonline (approved):

PR #629, opened 2025-10-07
fix #564: Sliders now start at correct default during onboarding

### LINKED ISSUE #564: Onboarding - Gift Compass
The sliders require some movement but it should start with the default value of 100% practical & purposeful, Cozy and Comfortable, or Simple & Essential if someone doesn't move it (because they don't need to).

Error message says "Expected number, received null"

tonyb650:
### Before: 
During the onboarding process, the "Gift Compass" sliders display at value 0 (all the way to the left) even though the default value for all 3 is set to "50". When the user attempts to progress to the next screen without adjusting the slider(s), an error is displayed "Expected number, received null"

### After: 
Sliders display default for new users correctly and the error no longer is displayed.

### Additional information:
The error was caused because user data fetched in `useEffect` was overwriting the default values being passed to the form, even when those values are `null`. I added `if` checks for each of the values, allowing the default form value to be overwritten only if there is a truthy value in the fetched data. Another, more concise, syntax would be e.g.: `form.setValue('giftStyle', data.minimal_luxurious || DEFAULT_FORM_VALUES.giftStyle)` but I think the explicit `if` check is a bit more clear.

Long term, perhaps validating the fetched user data with Zod or another technique would be a more robust solution.

### Also:
I made a comment of a possible grammatical improvement. If approved, this could potentially be included in this PR.

 Closes #564  

## Screenshot
gift_compass_screenshot

## Pre-submission checklist

### DIFF
--- a/app/onboarding/onboarding-form/onboarding-form.tsx
+++ b/app/onboarding/onboarding-form/onboarding-form.tsx
@@ -1,7 +1,7 @@
 // Copyright (c) Gridiron Survivor.
 // Licensed under the MIT License.
 
-import { useState, useEffect, JSX } from 'react';
+import { useState, useEffect, type JSX } from 'react';
 import { useRouter, useSearchParams } from 'next/navigation';
 import { z } from 'zod';
 import { useForm, Controller } from 'react-hook-form';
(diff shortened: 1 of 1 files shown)


tonyb650:
Ticket resolved through setting defaults upon record creation within Supabase.

PR #637, opened 2025-10-16
test #174: created unit test for Avatar component

### LINKED ISSUE #174: Create unit test for Avatar.tsx

jefische:
### Before: 
The `Avatar` component lacked unit tests.

#### Unit tests have been added for `Avatar` component:

- renders user avatar image when valid src attribute is provided
- renders fallback/default avatar image when src is undefined
- added alt text to `AvatarImage` to access element by semantic query

 Closes #174 

## Testing instructions

If applicable, provide steps for reviewers to test changes locally -- including necessary setup, commands, and expected results
 
## Additional information

Share any additional info that may provide context for the PR evaluation (performance considerations, design choices, etc)

## [optional] Screenshots

image

## Pre-submission checklist


components/Avatar/Avatar.test.tsx
```diff
@@ -0,0 +1,35 @@
+import Avatar from './Avatar';
+import { render, screen } from '@testing-library/react';
+
+jest.mock('@radix-ui/react-avatar', () => {
+  const actual = jest.requireActual('@radix-ui/react-avatar');
+  return ({
+    ...actual,
+    Image: ({src, alt, ...props}: {src?: string, alt?: string}) => {
```
suzuhe481: Remove `...props` from the Image mock. When testing, you should only mock the necessary behavior. For the `Avatar` component, you're currently only testing for `src` and `alt`. The other properties of Image aren't needed.

components/Avatar/Avatar.test.tsx
```diff
@@ -0,0 +1,35 @@
...
+      return <img data-testid="avatar-image" src={src} alt={alt} {...props}/>;
```
suzuhe481: Don't add data-testids inside components used in testing. While you fetched the element by role, which is good, the id still isn't necessary for the mock.

If anything, this id should have been added to the `AvatarImage` component.

components/Avatar/Avatar.tsx
```diff
@@ -19,7 +19,7 @@ const Avatar = ({
 }): JSX.Element => {
   return (
     <AvatarBody>
-      <AvatarImage src={userAvatar} alt="" />
+      <AvatarImage src={userAvatar} alt="user avatar" />
```
nickytonline: refactor (a11y): `"user avatar"` is too generic. If the avatar conveys meaning, add more context, like the username, otherwise leave the alt test to `""`.

```suggestion
      <AvatarImage src={userAvatar} alt=`profile image for user ${username}` />
```

components/Avatar/Avatar.test.tsx
```diff
@@ -0,0 +1,35 @@
+import Avatar from './Avatar';
+import { render, screen } from '@testing-library/react';
+
+jest.mock('@radix-ui/react-avatar', () => {
```
nickytonline: Just FYI for anyone coming across this. My first instinct was to ask why are we mocking this as you always want to mock as little as possible or not at all so we are testing the actual implementation.

In this case, we mock Radix’s Avatar.Image because in JSDOM, what we use in Jest tests, image load events don’t fire as it's not a real browser environment, so only the fallback is ever shown in tests if we don't mock.

nickytonline (changes requested):

components/Avatar/Avatar.tsx
shashilo: Also, this is only a unit test task. This should be removed into an A11Y ticket. I want to keep things simple for your first test.

components/Avatar/Avatar.test.tsx
```diff
@@ -0,0 +1,35 @@
...
+    const avatarElementWithImage = screen.getByRole('img', { name: 'user avatar'});
+    expect(avatarElementWithImage).toBeInTheDocument();
+    expect(avatarElementWithImage).toHaveAttribute(
+        'src',
+        'https://mdbcdn.b-cdn.net/img/new/avatars/2.webp',
+      );    
+
+  });
+
+  it('Renders the fallback image when src is undefined', () => {
```
shashilo: What about when there's a string given bug the image is not found?

shashilo (changes requested):

components/Avatar/Avatar.test.tsx
jefische: That makes sense, I'll add this test scenario.

components/Avatar/Avatar.test.tsx
```diff
@@ -0,0 +1,35 @@
...
+    ...actual,
+    Image: ({src, alt, ...props}: {src?: string, alt?: string}) => {
+      return <img data-testid="avatar-image" src={src} alt={alt} {...props}/>;
+    }
+  })
+});
+
+describe('Avatar Component', () => {
+
+  it('Renders the component with user image using valid src attribute', () => {
```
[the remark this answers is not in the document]
jefische: Not for this scenario. I wanted to test that the user provided image is correctly added to the document.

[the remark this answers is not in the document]
jefische: Interestingly the fallback image is still showing in the document for this scenario. I tried using `.not.toBeVisible()` method as well as `.not.toBeInTheDocument()` but both tests fail:

1st attempt:
```diff
+ const avatarElementFallbackImage = screen.getByTestId('avatar-fallback')
+ expect(avatarElementFallbackImage).not.toBeVisible();
```

2nd attempt:
```diff
+ const avatarElementFallbackImage = screen.getByTestId('avatar-fallback')
+ expect(avatarElementFallbackImage).not.toBeInTheDocument();
```
I tried looking at `screen.debug()` and it's showing both elements in the JSDOM. Are both of these supposed to show up like this?

Screen-debug-avatar-component

shashilo (dismissed):

nickytonline (dismissed):
🚀

PR #657, opened 2025-10-27
feat #655: implements an error handling system with custom errors

### LINKED ISSUE #655: Update Next.js Responses to Display Full Error Message
When errors are caught while making API calls, the thrown errors are sometimes not detailed enough or incorrect. The error that was caught should be what is thrown so it can be logged and viewed in Vercel's logs. This will allow us to see what the actual errors are.

### Example of Current Implementation
When catching an error, throw it's actual message instead of creating a custom one.
Image

suzuhe481:
### Before: 
Backend routes/functions were thrown or returned without always being caught.

### After: 
- Adds custom error classes to define specific error types with the file located in `lib/errors/CustomErrors.ts`.
  - `BackendError`, `OpenAiError`, `SupabaseError`, `GoogleError`
- Errors will only be logged at the top level of where a function/route was called. Will typically be right before a response is sent to the client. Normally, this is the catch block inside of a `route.ts`.
- Errors created/caught in nested functions will be continuously thrown until it gets caught in a catch block.
  - eg. The `generateAndStoreSuggestions()` function throws an `OpenAiError`, which gets rethrown in `drawGiftExchange()`, rethrown again in `processGiftExchanges()`, then finally caught and logged in a `route.ts`.
- Some functions didn't need to return anything, but still returned a `success: boolean`. I removed these returns, but errors will still be thrown.
- Added a `logError(error: unknown)` function located in `lib/errors/logError.ts`. This function is accepts the error to correctly identify and log it, as well as being easily expandable to new error types if they get created.

 Closes #655 

### Example of error log in the terminal
Error log example
 

## Pre-submission checklist

### DIFF
--- a/app/create-group-page/page.tsx
+++ b/app/create-group-page/page.tsx
@@ -23,7 +23,7 @@ import {
   PopoverContent,
   PopoverTrigger,
 } from '@/components/Popover/popover';
-import { format } from 'date-fns';
+import { addDays, format } from 'date-fns';
 import {
   Command,
   CommandEmpty,
@@ -85,8 +85,8 @@ export default function CreateGroupPage() {
     defaultValues: {
       name: '',
       description: '',
-      drawing_date: new Date(),
-      exchange_date: new Date(),
+      drawing_date: undefined,
+      exchange_date: undefined,
       budget: '',
       group_image: GROUP_IMAGES[0].src,
     },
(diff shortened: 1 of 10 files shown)


shashilo (changes requested):
Use the status code from the response and now one we give out. Each of the API's should have a status code property that comes along with it.

PR #661, opened 2025-10-29
fix #650: Fix GiftExchangePage crashes w/ non-group user and status is not pending

### LINKED ISSUE #650: Gift Exchange page issue that crashes
When a non-user/non-member tries to navigate to a gift exchange page AFTER it has already been drawn, the app crashes.

Image

tonyb650:
### Before: 
When a user, whether logged in or not, tries to navigate to a gift exchange page **after** the drawing has occurred (`status` is no longer 'pending'), the app crashes.

This occurred because the page should not be rendered at all if the group has already held it's drawing and the user is not part of that group.

### After: 
As part of the data fetching function that is fired by the page's `useEffect`, the following checks and new behaviors have been added:
1) If the `isSignedIn` status of the user has not been established yet, there is an early return before any data fetching
2) If the data fetching returns with an `error` property, the user is redirected to dashboard with a "Bad Link" toast (Case 1)
3) If the user is not logged in and the GiftExchange is no longer 'pending' (i.e.: the Drawing has already occurred), then the `signInWithGoogle` function is called (Case 2)
4) If the user is logged in but not a member of a Gift exchange that is no longer 'pending', then the user is redirect to dashboard with and "Expired Link" toast (Case 3)
5) If an error is caught during data fetching or parsing, then a "Server Error" toast is rendered.

### Additional
AuthContext was updated to make the behavior of `isSignedIn` more useful. `isSignedIn` now remains `null` until the session status can be determined and then is set to `true` or `false` once the session status is known.

 Closes #650

## Testing instructions
To test,
* Case 1: Create a user and complete onboard. Type/paste an 'invite' link with a bad `id` param into browser. This will render the "Bad Link" toast.
* Create a group or update a group and copy and save the 'invite' link.
* Complete the 'drawing' action. This will update the `status` to 'active'.
* Case 2. Log out. Paste in the link to the browser. This will initiate Google login.
* Case 3. Log in as a user who is not a part of the group. Paste in the link to the browser to get the 'Expired' toast.
 
## Screenshots
image

## Pre-submission checklist

### DIFF
--- a/app/create-group-page/page.test.tsx
+++ b/app/create-group-page/page.test.tsx
@@ -21,6 +21,38 @@ class MockResizeObserver {
 global.ResizeObserver = MockResizeObserver;
 
 describe('Create Group Page', () => {
+  it('has the first group image selected by default', () => {
+    render(<CreateGroupPage />);
+
+    const [firstTile, ...otherTiles] = screen.getAllByRole('figure');
+    expect(firstTile).toHaveAttribute('data-state', 'checked');
+    otherTiles.forEach((imageTile) => {
+      expect(imageTile).toHaveAttribute('data-state', 'unchecked');
+    });
+  });
+
+  describe('Close (X) button', () => {
+    it('renders the X button with the correct href', () => {
+      render(<CreateGroupPage />);
+
+      expect(screen.getByTestId('x-button')).toHaveAttribute(
+        'href',
+        '/dashboard',
+      );
+    });
+
+    describe('Cancel button', () => {
+      it('renders the Cancel button with the correct href', () => {
+        render(<CreateGroupPage />);
+
+        expect(screen.getByRole('link', { name: /cancel/i })).toHaveAttribute(
+          'href',
+          '/dashboard',
+        );
+      });
+    });
+  });
+
   describe('Calendar component in create group page', () => {
     const currentDate = new Date('2025-10-15T00:00:00Z');
 
@@ -52,16 +84,6 @@ describe('Create Group Page', () => {
 
       const tomorrow = screen.getByText('16');
       expect(tomorrow).not.toBeDisabled();
-    });
-  });
-
-  it('has the first group image selected by default', () => {
-    render(<CreateGroupPage />);
-
-    const [firstTile, ...otherTiles] = screen.getAllByRole('figure');
-    expect(firstTile).toHaveAttribute('data-state', 'checked');
-    otherTiles.forEach((imageTile) => {
-      expect(imageTile).toHaveAttribute('data-state', 'unchecked');
     });
   });
 });
--- a/app/create-group-page/page.tsx
+++ b/app/create-group-page/page.tsx
@@ -13,7 +13,7 @@ import {
   FormLabel,
   FormMessage,
 } from '@/components/Form/form';
-import { Check, ChevronsUpDown, ChevronLeft } from 'lucide-react';
+import { Check, ChevronsUpDown, ChevronLeft, X } from 'lucide-react';
 import { Input } from '@/components/Input/Input';
 import { Button } from '@/components/Button/button';
 import { CalendarIcon } from 'lucide-react';
@@ -39,6 +39,7 @@ import {
 } from '@/components/ImageSelector/ImageSelector';
 import { useRouter } from 'next/navigation';
 import LinkCustom from '@/components/LinkCustom/LinkCustom';
+import Link from 'next/link';
 
 const priceRanges = [
   { label: '$10 - $20', value: '10-20' },
@@ -121,7 +122,14 @@ export default function CreateGroupPage() {
       </div>
       <div className="flex items-center justify-center h-full">
         <div className="bg-white w-full xl:w-1/2 mb-5 flex justify-center align-center rounded flex-col ">
-          <h2 className="font-bold m-5">Create Secret Santa Page</h2>
+          <div className="flex justify-end m-1.5">
+            <Button variant="ghost" asChild data-testid="x-button">
+              <Link href="/dashboard">
+                <X className="text-black" />
+              </Link>
+            </Button>
+          </div>
+          <h2 className="font-bold mx-5 mb-5">Create Secret Santa Page</h2>
           <Form {...form}>
             <form onSubmit={form.handleSubmit(onSubmit)} className="space-y-8">
               <FormField
@@ -329,10 +337,11 @@ export default function CreateGroupPage() {
                   </FormItem>
                 )}
               />
-              <div className="flex justify-center md:justify-start md:m-5 m-0 w-full">
-                <Button className="m-2" type="submit">
-                  Create Group
+              <div className="flex gap-2 justify-center md:justify-end p-3">
+                <Button variant="secondary" className="bg-slate-300" asChild>
+                  <Link href="/dashboard">Cancel</Link>
                 </Button>
+                <Button type="submit">Create Group</Button>
               </div>
             </form>
           </Form>


context/AuthContextProvider.tsx
```diff
@@ -70,8 +70,8 @@ const AuthContextProvider: FC<{ children: ReactNode }> = ({ children }) => {
             data: { user },
           } = await supabase.auth.getUser();
           setUser(user);
-          setIsSignedIn(true);
         }
+        setIsSignedIn(!!session);
```
shashilo: This is a significant change. Why was this moved outside the if statement? Also, no unit tests were updated. That's bad. Either we don't have a unit test for this yet, or there's a failing build coming soon.

app/gift-exchanges/[id]/page.tsx
```diff
@@ -69,29 +80,60 @@ export default function GiftExchangePage() {
...
+        });
+        return;
+      }
+
+      if (isUserLoggedIn && !isExchangePending && !isUserAGroupMember) {
+        router.push('/dashboard');
+        toast({
+          variant: ToastVariants.Error,
+          title: `Expired Link`,
+          description: `Sorry, the specified link is no longer valid.`,
```
Comment (shashilo) [inline app/gift-exchanges/[id]/page.tsx]:
```suggestion
          description: `Sorry, this invitation is no longer valid.`,
```

app/gift-exchanges/[id]/page.tsx
```diff
@@ -129,9 +171,9 @@ export default function GiftExchangePage() {
             </section>
             <section className="flex flex-col">
               <h1 className="font-bold">Gift Suggestions</h1>
-              {giftSuggestions.length === 0 && <WaitingForSuggestions />}
+              {giftSuggestions?.length === 0 && <WaitingForSuggestions />}
 
-              {giftSuggestions.length !== 0 && (
+              {giftSuggestions?.length > 0 && (
```
Comment (shashilo) [inline app/gift-exchanges/[id]/page.tsx]: What is this fixing?

app/gift-exchanges/[id]/page.test.tsx
```diff
@@ -32,37 +24,96 @@ jest.mock('@/app/api/openaiConfig/config', () => ({
...
+    (global.fetch as jest.Mock)
+      .mockResolvedValueOnce({
+        json: async () => ({ error: 'Error' }),
+      })
+      .mockResolvedValueOnce({
+        json: async () => ({ error: 'Error' }),
+      })
+      .mockResolvedValueOnce({
+        json: async () => ({ error: 'Error' }),
+      });
```
Comment (shashilo) [inline app/gift-exchanges/[id]/page.test.tsx]: Why is there 3 copies of the same mock method?

app/gift-exchanges/[id]/page.test.tsx
```diff
@@ -84,4 +136,67 @@ describe('GiftExchangePage Warning Modal', () => {
...
+  it('displays a Server Error toast if data fetching throws an error', async () => {
+    (useAuthContext as jest.Mock).mockReturnValue(mockMemberSession);
+    mockRejectedFetch();
+
+    render(<GiftExchangePage />);
+
+    await waitFor(() => {
+      expect(mockToast).toHaveBeenCalledWith({
+        variant: 'error',
+        title: 'Server Error',
```
Comment (shashilo) [inline app/gift-exchanges/[id]/page.test.tsx]: Instead of making these toast strings, create these strings into consts since they are not being reused.

shashilo (changes requested):

app/gift-exchanges/[id]/page.tsx
```diff
@@ -69,29 +80,60 @@ export default function GiftExchangePage() {
...
+      if (!isExchangePending && !isUserLoggedIn) {
+        await signInWithGoogle({
+          redirectPath: window.location.pathname,
+        });
+        return;
+      }
+
+      if (isUserLoggedIn && !isExchangePending && !isUserAGroupMember) {
+        router.push('/dashboard');
+        toast({
```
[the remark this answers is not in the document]
Comment (tonyb650) [reply app/gift-exchanges/[id]/page.tsx]: I don't think it matters. I tested it both ways and in my environment I couldn't discern any difference.

I'm thinking that if the ToastProvider was not global, then potentially a Toast that is rendered first would be removed with the redirect and that would mean the Toast should come after. But that's not the case here as far as I can tell

app/gift-exchanges/[id]/page.test.tsx
Comment (tonyb650) [reply app/gift-exchanges/[id]/page.test.tsx]: I copied the pattern that was already in the test for the 3 successful data fetching and it seemed to work well for the 3 "error" fetch calls needed for the 'bad link' response.

app/gift-exchanges/[id]/page.test.tsx
Comment (tonyb650) [reply app/gift-exchanges/[id]/page.test.tsx]: Done

app/gift-exchanges/[id]/page.tsx
```diff
@@ -69,29 +80,60 @@ export default function GiftExchangePage() {
...
+          description: `Sorry, the specified link is no longer valid.`,
+        });
+        return;
       }
     } catch (error) {
       console.error('Error fetching data:', error);
+      toast({
+        variant: ToastVariants.Error,
+        title: `Server Error`,
+        description: `Sorry, there was a network error.`,
```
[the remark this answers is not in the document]
Comment (nickytonline) [reply app/gift-exchanges/[id]/page.tsx]: +1 to this. Always display friendly errors and log the actual errors.

app/gift-exchanges/[id]/page.test.tsx
Comment (tonyb650) [reply app/gift-exchanges/[id]/page.test.tsx]: It needs to get back a response (either actual data or {error: "message"}) for all 3 fetch calls. If any of the 3 fetches are rejected then an error is thrown and we end up in the `catch` block, which isn't what we want for this execution path.   

I think a big improvement would be this:
```
const mockFetchWithoutValidData = () => {
   (global.fetch as jest.Mock)
     .mockResolvedValue({
       json: async () => ({error: "Error"}),
    })
}
```

Then all the fetch calls get the same response, regardless of how many there are.

app/gift-exchanges/[id]/page.test.tsx
```diff
@@ -84,4 +136,67 @@ describe('GiftExchangePage Warning Modal', () => {
...
+    (useAuthContext as jest.Mock).mockReturnValue(mockNonMemberSession);
+    mockDataFetch('active');
+
+    render(<GiftExchangePage />);
+
+    await waitFor(() => {
+      expect(routerPush).toHaveBeenCalledWith('/dashboard');
+
+      expect(mockToast).toHaveBeenCalledWith({
+        variant: 'error',
```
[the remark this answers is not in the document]
Comment (shashilo) [reply app/gift-exchanges/[id]/page.test.tsx]: Yes, and the other strings as well

context/AuthContextProvider.tsx
shashilo: Pretty much, which scenario is this fixing? Is it causing regressions?

tonyb650: Here's the scenario: The user is logged in, but when our page is first mounted, `useAuthContext` fires its useEffect and that in turn awaits `supabase.auth.getSession()`. While it is waiting, both `session` and `isSignedIn` are reverted to their default of `null`. If we are in this state and we don't exit early from `fetchGfitExchangeData`, we will enter `signInWithGoogle()` on line 107 every time.

I don't see it causing any regressions. The only place in the codebase `isSignedIn` is used is on the home page where it is compared with `true` to redirect the user to '/dashboard'. The new code should never change when it is set to `true`, only that it can now also be set to `false` in addition to `null` (when `session` is definitively not present).

app/gift-exchanges/[id]/page.test.tsx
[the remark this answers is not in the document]
Comment (tonyb650) [reply app/gift-exchanges/[id]/page.test.tsx]: Just realized, maybe you meant something like this:

const TOASTS = { badLink : { variant: ..., title: "Bad Link, description: "Please..."}, expiredLink: {variant: ..., title: "Expired Link", description: "..."}, ...}

Should I implement that and export it to the test file?

shashilo (approved):

lillian-nguyen (approved):

PR #697, opened 2025-11-19
fix: Vibe coding on fixing the peroformance issues for gift suggestions

shashilo:
## Description
This is an attempt to improve the performance issues we're seeing when drawing gift suggestions.

## Pre-submission checklist

- [ ] Request reviews from the `Peer Code Reviewers` and `Senior+ Code Reviewers` groups
- [ ] Thread has been created in Discord and PR is linked in `gis-code-questions`

## Summary by CodeRabbit

* **Refactor**
  * Parallelized retrieval of match and suggestion data for faster responses.
  * Parallel batch updates for gift assignments with stricter error handling for reliability.
  * Bulk suggestion processing and parallel image fetching for higher throughput.
  * Upgraded AI model and standardized suggestion format for richer, more consistent recommendations.
* **Tests**
  * Updated tests to validate bulk insertion of generated suggestions.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/app/api/gift-exchanges/[id]/giftSuggestions/route.ts
+++ b/app/api/gift-exchanges/[id]/giftSuggestions/route.ts
@@ -27,11 +27,11 @@ export async function GET(
       throw new SupabaseError('User is not authenticated or exists', 500);
     }
 
-    // Get match with full profile info
-    const { data: match, error: matchError } = await supabase
-      .from('gift_exchange_members')
-      .select(
-        `
+    const [matchResult, suggestionsResult] = await Promise.all([
+      supabase
+        .from('gift_exchange_members')
+        .select(
+          `
         id,
         recipient_id,
         recipient:profiles!gift_exchange_members_recipient_id_profiles_fkey (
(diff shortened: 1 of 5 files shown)


app/api/gift-exchanges/[id]/giftSuggestions/route.ts
```diff
@@ -27,11 +27,11 @@ export async function GET(
       throw new SupabaseError('User is not authenticated or exists', 500);
     }
 
-    // Get match with full profile info
-    const { data: match, error: matchError } = await supabase
-      .from('gift_exchange_members')
-      .select(
-        `
+    const [matchResult, suggestionsResult] = await Promise.all([
```
Comment (nickytonline) [inline app/api/gift-exchanges/[id]/giftSuggestions/route.ts:30]: Do we want all of these to succeed? If so, `Promise.all` works. Otherwise, consider [Promise.allSettled](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled)

lib/generateAndStoreSuggestions.ts
```diff
@@ -99,39 +103,71 @@ export async function generateAndStoreSuggestions(
...
-        description: String(suggestion.description),
-        matchReasons: Array.isArray(suggestion.matchReasons)
-          ? suggestion.matchReasons.map(String)
-          : [],
-        matchScore: Number(suggestion.matchScore),
-        imageUrl: amazonData.imageUrl || null,
+    const parsedResponse: IGeneratedSuggestionRaw[] = ((): IGeneratedSuggestionRaw[] => {
+      const raw = JSON.parse(jsonContent) as unknown[];
+      return raw.map((item) => {
+        const obj = item as Record<string, unknown>;
```
nickytonline: refactor (nit): might as well just cast once. Runtime validation might make sense too. We don't have Zod, but you could use type guards e.g. 

```
function isSuggestionLike(item: unknown): item is Record<string, unknown> {
  return typeof item === 'object' && item !== null;
}
```


```suggestion
      const raw = JSON.parse(jsonContent) as Record<string, unknown>[];
      return raw.map((item) => {
        const obj = item; // probably don't need obj intermediary var since it was for casting
```

app/api/gift-exchanges/[id]/giftSuggestions/route.ts
Comment (shashilo) [reply app/api/gift-exchanges/[id]/giftSuggestions/route.ts:30]: I believe we want them all to succeed.

alexappleget (approved):

nickytonline (approved):
a nit and a question but good to go

